Issue #3, opened 2025-03-05
`Graph` and `Nodes`


willGraham01:
We need a way to represent DAGs in our codebase.

- [x] [`NetworkX`](https://networkx.org) is a strong candidate for an existing graphs library. We should probably find and examine some alternatives before we decide.

Currently we have decided that this representation needs to have the following.

- [x] Graphs store `Node`s as the, well, Nodes.
- [x] Graphs are as static as possible (which extends to the nodes)
- [ ] Graph should have a `expectation` method that returns the expectation of the random variable / distribution represented by the graph.
- [x] Graphs that have more than one "outcome" node (as below) throw user errors as we don't know what to do with them

## Nodes

Our nodes will essentially be thin wrappers on top of families of distributions. Once a graph is constructed, its nodes should remain as static as possible.

Each `Node` should have:

- [x] A name that can be set by the user (which we have intentions of using to match to data columns from data-files.
- [x] A collection of methods; `is_root`, `is_intermediary`, and `is_outcome`. A root node has no parents ("hidden causes"), an intermediary has at least one parent and one child, and an outcome has no children.
- [x] A node needs to know which `DistributionFamily` it describes.
- [x] A node needs a method to map from the parameters obtained from its parents, to the `Distribution` realised from these parameters and it's `DistributionFamily` (or alternatively, a direct sampling method given the number of samples requested and the parameters from its parents).
- [ ] A node might also need to know its basis functions (possibly only for outcome nodes).

PR #5, opened 2025-03-07
Add classes for storing graph

mscroggs:
Progress towards #3 

Add:

- `Node` protocol that defined the interface a node must have
- `RootDistributionNode` for nodes with no parents
- `DistributionNode` for nodes with parents
- placeholder classes `Distribution` and `DistributionFamily` to be replaced later
- `Graph` that wraps a networkx graph

I've opened #7 with ideas for how to make a nicer graph building interface on top of these data structures.

### DIFF
--- a/src/causalprog/graph/graph.py
+++ b/src/causalprog/graph/graph.py
@@ -3,24 +3,18 @@
 import networkx as nx
 from .node import Node
 
-g_index = 0
-
 
 class Graph(object):
     """A directed acyclic graph that represents a causality tree."""
 
-    def __init__(self, graph: nx.Graph, label: str | None = None):
+    def __init__(self, graph: nx.Graph, label: str) -> None:
         """Initialise a graph from a NetworkX graph."""
-        global g_index
+
         for node in graph.nodes:
             if not isinstance(node, Node):
-                raise ValueError(f"Invalid node: {node}")
+                raise TypeError(f"Invalid node: {node}")
 
-        if label is None:
-            self._label = f"Graph{g_index}"
-            g_index += 1
-        else:
-            self.label = label
+        self._label = label
 
         self._graph = graph
         self._nodes = list(graph.nodes())
--- a/src/causalprog/graph/graph.py
+++ b/src/causalprog/graph/graph.py
@@ -1,0 +1,27 @@
+"""Graph storage."""
+
+import networkx as nx
+from .node import Node
+
+
+class Graph(object):
+    """A directed acyclic graph that represents a causality tree."""
+
+    def __init__(self, graph: nx.Graph):
+        """Initialise a graph from a NetworkX graph."""
+        for node in graph.nodes:
+            if not isinstance(node, Node):
+                raise ValueError(f"Invalid node: {node}")
+
+        self._graph = graph
+        self._nodes = list(graph.nodes())
+        self._depth_first_nodes = list(nx.algorithms.dfs_postorder_nodes(graph))
+
+        self._predecessors = nx.algorithms.dfs_predecessors(graph)
+        self._successors = nx.algorithms.dfs_successors(graph)
+
+        for n in self._nodes:
+            if n not in self._predecessors:
+                self._predecessors[n] = []
+            if n not in self._successors:
+                self._successors[n] = []
--- a/src/causalprog/graph/graph.py
+++ b/src/causalprog/graph/graph.py
@@ -17,18 +17,9 @@ class Graph:
 
         self._label = label
 
-        self._graph = graph
+        self._graph = graph.copy()
         self._nodes = list(graph.nodes())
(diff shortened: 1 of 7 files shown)


src/causalprog/graph/graph.py
```diff
@@ -0,0 +1,45 @@
...
+                raise TypeError(msg)
+
+        self._label = label
+
+        self._graph = graph
+        self._nodes = list(graph.nodes())
+        self._depth_first_nodes = list(nx.algorithms.dfs_postorder_nodes(graph))
+
+        self._predecessors = nx.algorithms.dfs_predecessors(graph)
+        self._successors = nx.algorithms.dfs_successors(graph)
```
willGraham01, line 25: Do we want to take a copy of `graph` here, just in case the user decides to edit the `nx.Graph` object in their script / analysis later and doesn't appreciate the consequences of doing so?

Setting `self._predecessors` (and the others attributes) implies that we want to be able to treat `self._graph` as immutable. But because Python is Python, there is still scope for errors due to changing the underlying objects.

There might also be a case for making `_predecessors` (etc) into `properties`, but I'm guessing they'd rather be attributes so we can set them once and don't have to re-run the `dfs` algorithm over and over again.

src/causalprog/graph/graph.py
```diff
@@ -0,0 +1,45 @@
...
+                self._predecessors[n] = []
+            if n not in self._successors:
+                self._successors[n] = []
+
+        outcomes = [node for node in self._nodes if node.is_outcome]
+        if len(outcomes) == 0:
+            msg = "Cannot create graph with no outcome nodes"
+            raise ValueError(msg)
+        if len(outcomes) > 1:
+            msg = "Cannot yes create graph with multiple outcome nodes"
```
willGraham01, line 38:
```suggestion
            msg = "Cannot yet create graph with multiple outcome nodes"
```

src/causalprog/graph/node.py
```diff
@@ -0,0 +1,115 @@
...
+
+class Distribution:
+    """Placeholder class."""
+
+
+@runtime_checkable
+class Node(Protocol):
+    """An abstract node in a graph."""
+
+    @abstractproperty
```
willGraham01, line 21: Going to X-reference to here <https://github.com/UCL/causalprog/pull/8#discussion_r1993082197>

src/causalprog/graph/node.py
```diff
@@ -0,0 +1,115 @@
...
+
+    @property
+    def is_outcome(self) -> bool:
+        """Identify if the node is an outcome."""
+        return self._outcome
+
+    @property
+    def is_intermediary(self) -> bool:
+        """Identify if the node is an intermediary."""
+        return False
```
willGraham01, line 75:
```suggestion
        return not (self.is_root or self.is_outcome)
```

Right? A node is an intermediary node IFF it has at least one successor and one predecessor.

Though down on [line 115](https://github.com/UCL/causalprog/pull/5/files#diff-3c3e22e486429270877f6243bcb4b4ef3e5c25ae5e72e0391292b25486c23d09R115) it looks like it's an intermediary only if it has at least one parent?

tests/test_graph.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_simple_graph() -> None:
+    """Test a simple graph."""
+    family = causalprog.graph.node.DistributionFamily()
+    n_x = causalprog.graph.RootDistributionNode(family, "N_X")
+    n_m = causalprog.graph.RootDistributionNode(family, "N_M")
+    u_y = causalprog.graph.RootDistributionNode(family, "U_Y")
+    x = causalprog.graph.DistributionNode(family, "X")
+    m = causalprog.graph.DistributionNode(family, "M")
+    y = causalprog.graph.DistributionNode(family, "Y", is_outcome=True)
```
willGraham01, line 36: We probably want to bundle this guy up into a fixture at some point... I imagine we will be using `simple_graph` in a lot of tests going forward!

tests/test_graph.py
```diff
@@ -0,0 +1,43 @@
...
+
+    nx_graph = nx.Graph()
+    nx_graph.add_edges_from([[n_x, x], [n_m, m], [u_y, y], [x, m], [m, y]])
+
+    graph = causalprog.graph.Graph(nx_graph, "G0")
+
+    assert graph.label == "G0"
```
willGraham01, line 43: I guess we should `assert` the connectivity that we added has taken root? IE maybe check `predecessors` and `successors` for some of the Nodes?

willGraham01 (commented):
Some minor decision points raised (they might not need changing, just me asking), but looks good to go otherwise.

Might want to refactor the simple graph into a test fixture since I imagine we'll want it in many places further down the line. And with it being a fixture, it can also be given a docstring to describe (or draw) the graph itself for ease.

willGraham01 (approved):

PR #6, opened 2025-03-07
allow unused imports in __init__ files

mscroggs:
Having to put `# noqa: F401` by all unused imports in __init__.py feels over the top to me


matt-graham (approved):
Agreed that this seems a sensible global ignore to have

willGraham01 (approved):

Issue #4, opened 2025-03-05
Distributions and Distribution Families


willGraham01:
### Is Your Feature Request Related to a Problem? Please Describe

-

### Describe the Solution You'd Like

We make a distinction between families of distributions, and realised distributions, using the `DistributionFamily` and `Distribution` classes.

## Families

`DistributionFamily`s store the functional form of their members, given the parameters (which might be a Jax function, EG). A base class will just expose this as a partial method / not-implemented method that subclasses must implement.

- [ ] They will need a `.construct` method which takes a set of parameter values and returns a `Distribution`. This will need to be a callable itself, when implemented it might be EG a Jax function.
- [ ] They should probably have a user-setable name to help describe what they are!

Essentially, a distribution family can map from a requested amount of samples AND set of parameters, to some sample values.

## Distributions themselves

`Distribution`s are realisations of one member of a family - IE a family + a fixed set of parameters. They allow sampling as well as other standard inquiry methods like "mean", "median", "moment", etc.

- [ ] Again, it might want a name
- [ ] Has a `.sample` method.

matt-graham:
As an additional breadcrumb, another JAX based library we may want to look at is [Oryx](https://www.tensorflow.org/probability/oryx/). High-level description from that page: _a library for probabilistic programming and deep learning built on top of JAX_. It includes a distributions module with various predefined distributions (I think generally wrapping corresponding objects in TensorFlow Probability). It also has support for performing various transformations on probabilistic programmings such as producing functions to sample, evaluate log probability density, intervene by setting values of certain variables.

PR #8, opened 2025-03-10
Distribution class (and backend-agnosticity)

willGraham01:
Related to #4 - attempts to provide the skeleton for the `Distribution` infrastructure.

Started at `Distributions` because they are less "meta" than the `DistributionFamilies`.

PR adds:

- A helper class that called `Translator` in order to remain backend-agnostic, at least at a high level. The `Translator` is my "quick hack" at remaining backend-agnostic.
  - In theory, users can subclass from `Distribution` (or simply create instances from) in order to use whatever (compatible) sampling backend that they choose. 
  - For us as developers, we can just subclass from `Distribution` directly to provide more concrete standard distributions using our chosen backend. We might have a `distraxTranslator` for example.
- Adds the `Distribution` class. It is instantiated by some object that supports a `.sample` method to again allow a user to "choose" their own backend if necessary.
  - `test_different_backends.py` demonstrates how this general `Distribution` class can be instantiated using either `distrax` or `NumPyro` as the backend.


src/causalprog/utils/translator.py
```diff
@@ -0,0 +1,141 @@
...
+    no guarantee that the signature of ``do_something`` maps identically to that of
+    ``obj.backend_method``. A ``Translator`` allows us to encode a mapping of
+    ``obj.backend_method``s arguments to those of ``do_something``.
+    """
+
+    backend_method: str
+    corresponding_backend_arg: dict[str, str]
+
+    @property
+    @abstractmethod
```
mscroggs, line 44: With added import above, this can be simplified to:

```suggestion
    @abstractproperty
```

src/causalprog/utils/translator.py
```diff
@@ -0,0 +1,141 @@
...
+Our frontend (or user-facing) classes each use a syntax that applies across the package
+codebase. By contrast, the various backends that we want to support will have different
+syntaxes and call signatures for the functions that we want to support. As such, we need
+a helper class that can store this "translation" information, allowing the user to
+interact with the package in a standard way but also allowing them to choose their own
+backend if desired.
+"""
+
+import inspect
+from abc import ABC, abstractmethod
```
mscroggs, line 13:
```suggestion
from abc import ABC, abstractmethod, abstractproperty
```

src/causalprog/utils/translator.py
```diff
@@ -0,0 +1,141 @@
...
+    corresponding_backend_arg: dict[str, str]
+
+    @property
+    @abstractmethod
+    def _frontend_method(self) -> str:
+        """Name of the frontend method that the backend is to be translated into."""
+        return ""
+
+    @property
+    @abstractmethod
```
mscroggs, line 50:
```suggestion
    @abstractproperty
```

src/causalprog/utils/translator.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+
+    backend_method: str
+    corresponding_backend_arg: dict[str, str]
+
+    @property
+    @abstractmethod
+    def _frontend_method(self) -> str:
+        """Name of the frontend method that the backend is to be translated into."""
+        return ""
```
mscroggs, line 47: Abstract methods and properties do not need to return anything. We could simplify this to:

```suggestion
        """Name of the frontend method that the backend is to be translated into."""
```

src/causalprog/utils/translator.py
```diff
@@ -0,0 +1,141 @@
...
+
+    @property
+    @abstractmethod
+    def compulsory_frontend_args(self) -> set[str]:
+        """Arguments that are required by the frontend function."""
+        return set()
```
mscroggs, line 53: ... also here
```suggestion
        """Arguments that are required by the frontend function."""
```

mscroggs (commented):
A few minor comments. Otherwise looks ready to merge

src/causalprog/utils/translator.py
willGraham01: `@abstractproperty` [is deprecated apparently](https://docs.python.org/3/library/abc.html#abc.abstractproperty), and they recommend combining `@property` and `@abstractmethod` (though the order of `property` and `abstractmethod` does matter, apparently).

mscroggs: that feels needlessly verbose but good to know

PR #9, opened 2025-03-10
Ruff & mypy linting rules.

willGraham01:
Adds additional rules to the ignore list as per internal Slack suggestions:

- `D105` missing docstrings on magic (dunder) methods - can potentially see the value in having such docstrings in some cases but don't think it should be required as typically their behaviour is obvious and function signature (and so interpretation of arguments) fixed by interface they are implementing
- `ANN` in tests
- `D` rules in tests

Also have added `exclude = ["tests"]` to the `tool.mypy` config section in the `pyproject.toml`, for the same reasons as `ANN` has been ignored.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,9 +1,6 @@
 [build-system]
 build-backend = "setuptools.build_meta"
-requires = [
-    "setuptools",
-    "setuptools-scm",
-]
+requires = ["setuptools", "setuptools-scm"]
 
 [project]
 authors = [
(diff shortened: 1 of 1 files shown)


willGraham01:
Am not sure why `pre-commit` has decided to compactify the lines in `pyproject.toml`, hence why this "looks" like a negative PR.

mscroggs (approved):

matt-graham (approved):
Looks good to me. Also not sure what is going on with the reformatting - I think it's probably the hook for `prettier` causing the reformatting but why it's suddenly decided to do this I don't know as I've more often had it reformat lists in the opposite more verbose direction. If causes issues / starts flip flopping recommended reformatting changes we can review it's configuration.

Issue #29, opened 2025-03-20
Multidimensional sampling


mscroggs:
The current development has focussed on single dimensional variables. There are places where we may have issues when using multi-dimensional variables, eg:

https://github.com/UCL/causalprog/blob/8c7a517392639910569a7d8e0f739b732b15e07a/src/causalprog/graph/node.py#L80

Issue #26, opened 2025-03-20
remove for loop over samples


mscroggs:
In #30, we added a for loop over samples:

https://github.com/UCL/causalprog/blob/8c7a517392639910569a7d8e0f739b732b15e07a/src/causalprog/graph/node.py#L78

This is slow. We need to look into ways to take all these samples at once

Issue #24, opened 2025-03-20
Remove jax hardcoding


mscroggs:
#30 added a bit of hardcoding of the use of jax. This should be made more general.

See also #27 and #28.

PR #30, opened 2025-03-20
Make distributions and graph talk to each other

mscroggs:
Builds on top of #16.

Removes placeholder distribution classes in graph, and plugs in the proper distribution classes in their place


src/causalprog/algorithms/expectation.py
```diff
@@ -1,15 +1,17 @@
...
 
 
 def sample(
     graph: Graph,
-    outcome_node_label: str | None = None,
-    samples: int = 1000,
+    outcome_node_label: str,
+    samples: int,
+    *,
+    rng_key: jax.Array,
```
mscroggs, line 14: See #24

src/causalprog/graph/node.py
```diff
@@ -72,20 +44,46 @@ class DistributionNode(Node):
...
+        output = np.zeros(samples)
+        new_key = jax.random.split(rng_key, samples)
+        for sample in range(samples):
+            parameters = {
+                i: sampled_dependencies[j][sample] for i, j in self._parameters.items()
+            }
+            concrete_dist = self._dist.construct(
+                **parameters, **self._constant_parameters
+            )
+            output[sample] = concrete_dist.sample(new_key[sample], 1)[0][0]
```
mscroggs, line 85: #25

src/causalprog/graph/node.py
```diff
@@ -72,20 +44,46 @@ class DistributionNode(Node):
...
+        rng_key: jax.Array,
+    ) -> npt.NDArray[float]:
         """Sample a value from the node."""
-        return self._dist.sample(sampled_dependencies, samples)
+        if not self._parameters:
+            concrete_dist = self._dist.construct(**self._constant_parameters)
+            return concrete_dist.sample(rng_key, samples)
+        output = np.zeros(samples)
+        new_key = jax.random.split(rng_key, samples)
+        for sample in range(samples):
```
mscroggs, line 78: #26

src/causalprog/graph/node.py
```diff
@@ -72,20 +44,46 @@ class DistributionNode(Node):
...
+            parameters = {
+                i: sampled_dependencies[j][sample] for i, j in self._parameters.items()
```
mscroggs, line 80: #29

src/causalprog/distribution/normal.py
```diff
@@ -76,7 +78,7 @@ def __init__(self) -> None:
         """Create a family of normal distributions."""
         super().__init__(Normal, family_name="Normal")
 
-    def construct(self, mean: ArrayCompatible, cov: ArrayCompatible) -> Normal:
+    def construct(self, mean: ArrayCompatible, cov: ArrayCompatible) -> Normal:  # type: ignore # noqa: PGH003
```
mscroggs, line 81: mypy was very unhappy with everything I tried putting here. @willGraham01: Any idea what it should be?

mscroggs: Also, we originally had this as the following, but either ruff or mypy was not happy that the base class didn't have the `*,`, but having `*, **parameters` in the base class is a syntax error...

```suggestion
    def construct(self, *, mean: ArrayCompatible, cov: ArrayCompatible) -> Normal:  # type: ignore # noqa: PGH003
```

src/causalprog/_abc/labelled.py
```diff
@@ -20,4 +20,4 @@ def label(self) -> str:
         return self._label
 
     def __init__(self, *, label: str) -> None:
-        self._label = str(label)
+        self._label = label
```
willGraham01, line 23:
```suggestion
        self._label = str(label)
```

Looks to have snuck in again?

src/causalprog/distribution/normal.py
willGraham01: I think this is `mypy` expects any overwritten methods to retain a signature that is compatible with the signature in the base class. Personally I would have also thought that `*, mean, cov` was compatible with `**parameters` (since they're all keyword arguments) but in hindsight I can see why it's not (since `**parameters` does not necessitate ANY arguments, but `*, mean, cov` necessitates 2).

I think (for now), leave things as they are (`*, mean, cov`). I can tackle this when I re-work the `DistributionFamily`s using the backend-agnostic classes (which should make this awkward hand-me-down of parameters redundant).

src/causalprog/graph/graph.py
```diff
@@ -50,10 +50,12 @@ def add_edge(self, first_node: Node | str, second_node: Node | str) -> None:
...
-        for node_to_check in (first_node, second_node):
-            if node_to_check != self._nodes_by_label[node_to_check.label]:
-                msg = "Invalid node: {node_to_check}"
-                raise ValueError(msg)
+        if first_node != self._nodes_by_label[first_node.label]:
+            msg = "Invalid node"
+            raise ValueError(msg)
+        if second_node != self._nodes_by_label[second_node.label]:
+            msg = "Invalid node"
+            raise ValueError(msg)
```
willGraham01, line 58:
```suggestion
        for node_to_check in (first_node, second_node):
            if node_to_check != self._nodes_by_label[node_to_check.label]:
                msg = "Invalid node: {node_to_check}"
                raise ValueError(msg)
```

Looks like you're loosing some of the changes that were made to #16 in the review comments, before we started this during pair-programming. This shouldn't be changing back.

src/causalprog/graph/node.py
```diff
@@ -72,20 +44,46 @@ class DistributionNode(Node):
...
+        parameters: dict[str, str] | None = None,
+        constant_parameters: dict[str, float] | None = None,
         is_outcome: bool = False,
     ) -> None:
         """Initialise."""
         self._dist = distribution
+        if constant_parameters is None:
+            self._constant_parameters = {}
+        else:
+            self._constant_parameters = constant_parameters
```
willGraham01, line 59:
```suggestion
        self._constant_parameters = constant_parameters if constant_parameters else {}
```

src/causalprog/graph/node.py
```diff
@@ -72,20 +44,46 @@ class DistributionNode(Node):
...
+        if parameters is None:
+            self._parameters = {}
+        else:
+            self._parameters = parameters
```
willGraham01, line 63:
```suggestion
        self._parameters = parameters if parameters else {}
```

tests/test_distributions/test_family.py
```diff
@@ -2,7 +2,7 @@
 import pytest
 
 from causalprog.distribution.base import SampleTranslator
-from causalprog.distribution.family import DistributionFamily
+from causalprog.distribution.normal import DistributionFamily
```
willGraham01, line 5:
```suggestion
from causalprog.distribution.family import DistributionFamily
```

Not sure why it's now importing from a different module.

tests/test_graph.py
```diff
@@ -113,27 +116,58 @@ def test_cycle() -> None:
...
+        "X",
+        constant_parameters={"mean": mean, "cov": stdev**2},
+        is_outcome=True,
+    )
 
-    graph = causalprog.graph.Graph("G0")
+    graph = Graph("G0")
     graph.add_node(node)
 
+    key = jax.random.split(rng_key, 1)[0]
```
willGraham01, line 136:
```suggestion
    # To compensate for rng-key splitting in sample methods, note the "split" key
    # that is actually used to draw the samples from the distribution, so we can
    # attempt to replicate its behaviour explicitly.
    key = jax.random.split(rng_key, 1)[0]
```

Just adding this comment here as between yesterday morning and now I had forgotten why we did this :sweat_smile:

willGraham01 (approved):
Looks like a few changes from #16 are being re-overwritten here, but otherwise just a couple of shortenings.

With regards to the `construct` problems, I'll tackle them in my backend-agnostic PR anyway so happy to postpone with the `# noqa` until then.

tests/test_distributions/test_family.py
mscroggs: I think it's nicer to import from family as that's where this is defined?

Issue #19, opened 2025-03-18
Ruff exclusions


willGraham01:
Catch-all issue to tag when opening a pull request that edits the Ruff configuration.

When the project started, we adopted all `Ruff` & `mypy` rules, which we might not want to adhere to going forward (since it makes the code structure very strict, and some rules are more guidelines / style).

If you want to propose a rule change / drop / addition, open a pull request and mention (but do not close) this issue.

willGraham01:
`PLR2004` I think is generally useful (as magic numbers are bad). Though in the example given, there's no harm in a `# noqa: PLR2004`.

`PLR0913`: Yeah that one can definitely go.

willGraham01:
- [ ] `PIE807` should be excluded from the tests. We'll need to "delay evaluation" in several places, and this throws a warning whenever we do.

PR #32, opened 2025-03-20
ruff: skip PLR0913 in tests/

mscroggs:
See #19


mscroggs:
@willGraham01 do we want to skip this everywhere? Or just in tests?

(Currently I've done just in tests)

willGraham01 (approved):
Let's stick to just the tests for now. On the user-facing side it's probably a good thing to try to limit ourselves (for the time being!)

PR #39, opened 2025-04-29
add parameter node

### LINKED ISSUE #36: Add parameter node
A node that cannot be sampled but the value can be provided so that dependencies can be sampled

mscroggs:
Resolves #36

### DIFF
--- a/src/causalprog/graph/node.py
+++ b/src/causalprog/graph/node.py
@@ -115,7 +115,7 @@ class ParameterNode(Node):
         if self.value is None:
             msg = "Cannot sample an undetermined parameter node."
             raise ValueError(msg)
-        return np.full(self.value, samples)
+        return np.full(samples, self.value)
 
     def __repr__(self) -> str:
         return f'ParameterNode("{self.label}")'
--- a/src/causalprog/graph/node.py
+++ b/src/causalprog/graph/node.py
@@ -99,7 +99,7 @@ class ParameterNode(Node):
     """A node containing a parameter."""
 
     def __init__(
-        self, label: str, *, value: int | None = None, is_outcome: bool = False
+        self, label: str, *, value: float | None = None, is_outcome: bool = False
     ) -> None:
         """Initialise."""
         super().__init__(label, is_outcome=is_outcome, is_parameter=True)
--- a/src/causalprog/graph/node.py
+++ b/src/causalprog/graph/node.py
@@ -98,18 +98,18 @@ class DistributionNode(Node):
 class ParameterNode(Node):
     """
     A node containing a parameter.
-    
+
     `ParameterNode`s differ from `DistributionNode`s in that they do not have an attached distribution
     (family), but rather represent a parameter that contributes to the shape of one (or more)
     `DistributionNode`s.
-    
+
     The collection of parameters described by `ParameterNode`s forms the set of variables that will
     be optimised over in the corresponding `CausalProblem`. `ParameterNode`s have a `.value` attribute
     which stores the current value of the parameter to facilitate this - a `CausalProblem` needs
     to be able to update the values of the parameters so it can make evaluations of the causal estimand
     and constraints functions, _as if_ they were functions of the parameters, rather than the
     `DistributionNode`s.
-    
+
     `ParameterNode`s should not be used to encode constant values used by `DistributionNode`s.
     Such constant values should be given to the necessary `DistributionNode`s directly as
     `constant_parameters`.
--- a/src/causalprog/graph/node.py
+++ b/src/causalprog/graph/node.py
@@ -130,7 +130,7 @@ class ParameterNode(Node):
     ) -> npt.NDArray[float]:
         """Sample a value from the node."""
         if self.value is None:
-            msg = "Cannot sample an undetermined parameter node."
+            msg = f"Cannot sample undetermined parameter node: {self.label}."
             raise ValueError(msg)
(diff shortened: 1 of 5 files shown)


src/causalprog/graph/node.py
```diff
@@ -81,3 +93,26 @@ def sample(
...
+class ParameterNode(Node):
+    """A node containing a parameter."""
+
+    def __init__(self, label: str, *, is_outcome: bool = False) -> None:
+        """Initialise."""
+        super().__init__(label, is_outcome=is_outcome, is_parameter=True)
+
+    def sample(
+        self,
+        sampled_dependencies: dict[str, npt.NDArray[float]],
```
mscroggs, line 107:
```suggestion
        _sampled_dependencies: dict[str, npt.NDArray[float]],
```

src/causalprog/graph/node.py
```diff
@@ -81,3 +93,26 @@ def sample(
...
+        _samples: int,
+        _rng_key: jax.Array,
+    ) -> npt.NDArray[float]:
+        """Sample a value from the node."""
+        if self.label not in sampled_dependencies:
+            msg = "Cannot sample an undetermined parameter node."
+            raise ValueError(msg)
+        return sampled_dependencies[self.label]
```
mscroggs, line 115:
```suggestion
        if self.value is None:
            msg = "Cannot sample an undetermined parameter node."
            raise ValueError(msg)
        return np.full(self.value, samples)
```

src/causalprog/graph/node.py
```diff
@@ -81,3 +93,26 @@ def sample(
...
     def __repr__(self) -> str:
         return f'DistributionNode("{self.label}")'
+
+
+class ParameterNode(Node):
+    """A node containing a parameter."""
+
+    def __init__(self, label: str, *, is_outcome: bool = False) -> None:
+        """Initialise."""
+        super().__init__(label, is_outcome=is_outcome, is_parameter=True)
```
mscroggs, line 103:
```suggestion
        super().__init__(label, is_outcome=is_outcome, is_parameter=True)
        self.value: int | None = None
```

src/causalprog/graph/node.py
```diff
@@ -81,3 +93,27 @@ def sample(
...
+        self.value: int | None = None
```
mscroggs, line 104:
```suggestion
    def __init__(self, label: str, value: int | None, *, is_outcome: bool = False) -> None:
        """Initialise."""
        super().__init__(label, is_outcome=is_outcome, is_parameter=True)
        self.value = value
```

src/causalprog/graph/node.py
```diff
@@ -81,3 +93,29 @@ def sample(
 
     def __repr__(self) -> str:
         return f'DistributionNode("{self.label}")'
+
+
+class ParameterNode(Node):
+    """A node containing a parameter."""
```
willGraham01, line 99:
```suggestion
    """
    A node containing a parameter.
    
    `ParameterNode`s differ from `DistributionNode`s in that they do not have an attached distribution
    (family), but rather represent a parameter that contributes to the shape of one (or more)
    `DistributionNode`s.
    
    The collection of parameters described by `ParameterNode`s forms the set of variables that will
    be optimised over in the corresponding `CausalProblem`. `ParameterNode`s have a `.value` attribute
    which stores the current value of the parameter to facilitate this - a `CausalProblem` needs
    to be able to update the values of the parameters so it can make evaluations of the causal estimand
    and constraints functions, _as if_ they were functions of the parameters, rather than the
    `DistributionNode`s.
    
    `ParameterNode`s should not be used to encode constant values used by `DistributionNode`s.
    Such constant values should be given to the necessary `DistributionNode`s directly as
    `constant_parameters`.
    """
```

Just fleshing this out a little bit to include some of the discussion points we hit on Tuesday too.

src/causalprog/graph/node.py
```diff
@@ -81,3 +93,29 @@ def sample(
...
+
+    def sample(
+        self,
+        sampled_dependencies: dict[str, npt.NDArray[float]],  # noqa: ARG002
+        samples: int,
+        rng_key: jax.Array,  # noqa: ARG002
+    ) -> npt.NDArray[float]:
+        """Sample a value from the node."""
+        if self.value is None:
+            msg = "Cannot sample an undetermined parameter node."
```
willGraham01, line 116:
```suggestion
            msg = f"Cannot sample undetermined parameter node: {self.label}."
```

Likely will be helpful to us when debugging to know which node is causing issues :sweat_smile:

src/causalprog/graph/node.py
```diff
@@ -81,3 +93,29 @@ def sample(
...
+            raise ValueError(msg)
+        return np.full(samples, self.value)
```
willGraham01, line 118: Do we want to return a `jax.numpy.full` here? I guess right now it's OK as a `numpy` return since that's what our other algorithms are returning too, but we should probably move to using `jax.numpy` everywhere for future proofing.

willGraham01 (approved):
:+1: Nothing warranting functional changes. Though some questions about `jax` vs `numpy` - don't worry about addressing here but if you agree I'll open an issue for us so we don't forget to address the `jax.numpy <- numpy` issue later.

src/causalprog/graph/node.py
willGraham01: https://github.com/UCL/causalprog/issues/44

Issue #37, opened 2025-04-23
Graph can translate functions of the random variables to that of the probablistic parameters


willGraham01:
The graph class can identify which of its nodes are (probabalistic) parameters (PPs) thanks to #36. This means that the graph is effectively a map between these parameters and the random variables (RVs) that the problem concerns.

The causal estimand and constraint functions are typically expressed as functions of the RVs in the literature, and this is an approach we have decided to mimic in our package. However, since we will need to optimise over the PPs, the package needs to be able to translate functions of the RVs into functions of the PPs using the graph structure automatically. These new functions can then be passed into whatever optimisier we are using to solve for the upper and lower bounds.

As such, we need either a `Graph` method or `algorithm` function that translates a function of the RVs into a function of the PPs.

willGraham01:
Let's try and make the issue concrete. Take the setup of the toy example from #34, and let's say someone has written the function `sigma(rv_1) = expectation(rv_1)` (or `rv_1.expectation`), which is their causal estimand that acts on a RV `rv_1`. In the current setup, we are proposing that they then do 

```python
sigma_pp = graph.translate_to_pp_function(sigma, {"rv_1": "Y"})
```

to obtain a function `sigma_pp` that is now a function of $\mu_X$, which returns the expectation of `Y`. 

### The Problem

However, it is not as to create such a function as we expect, since the "composition" of `sigma` with the graph (acting as a map from the PP to the RVs) is not identical to what we expect `sigma_pp` to do:

- `sigma_pp` _should_ draw samples $x_i$ from $X$ using the fixed value of $\mu_X$, then for each $x_i$ construct a realisation $Y_i = \mathcal{N}(x_i, \nu_Y^2)$ of $Y$ and take _one_ sample from this, giving a series of samples $y_i$. The average of the $y_i$ is then returned as the expectation.
- `sigma` composed with the graph on the other hand does something different. It would draw _one_ sample $x_0$ from $X$, then use $x_0$ to construct a realisation $Y_0 = \mathcal{N}(x_0, \nu_Y^2)$ of $Y$. It would then compute `expectation(Y_0)`.

Crucially, `expectation(Y_0)` is _not the same_ as the result of averaging the $y_i$s (I don't think it even is the same for normal distributions like in our toy problem, and it certainly is not for other distributions!).

[...]

PR #40, opened 2025-04-29
`CausalEstimand` class for evaluating causal estimands

willGraham01:
Partially addresses #37 | Target is #39 since `ParameterNode`s are needed to get this functionality working, but @mscroggs feel free to hold off the merge until you're happy with the base (since this will need rebasing after any updates to that branch, anyway).

Introduces the `CausalProblem` class, designed to handle interactions between a `Graph` (`.graph`), the causal estimand of interest (`.causal_estimand`) and the constraints (currently not implemented). Handing of the constraints is not introduced in this pull request, since any comments about how the `causal_estimand` is being handled would apply to the constraints too, so figured it was best to settle on a design for one before writing the next.

# Key Changes

## Graph class

Had to introduce two methods related to `ParameterNode`s - mostly for fetching a tuple of all such nodes (note that this method uses `graph.ordered_nodes` since it is important that the parameter nodes a consistently given in the same order) and for the setting of the `.value` attributes of these nodes.

@mscroggs you might want to steal [the corresponding changes](https://github.com/UCL/causalprog/pull/40/files#diff-baee8e6f9e5532715e4174912a1264666f3563832b010c7a8cd381a875d91ecbR59) & [their tests](https://github.com/UCL/causalprog/pull/40/files#diff-e9352f44736d5b9e922852f07917f4d93bb6d4ca2e6d9743abf20194a5c414d5) and move them into the base PR (#39).

## Node class

The `sample` method for `ParameterNode`s [needs to have the same argument names](https://github.com/UCL/causalprog/pull/40/files#diff-3c3e22e486429270877f6243bcb4b4ef3e5c25ae5e72e0391292b25486c23d09) as the super-class for the "up-the-graph" logic to work without an error, so have removed the `_` prefixes and slapped on a `# noqa` for now, until we find a more elegant solution.

I also took the [liberty of telling the linter](https://github.com/UCL/causalprog/pull/40/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R83) that unused `*args` and `**kwargs` are OK. Since we're going to have a lot of inherited and overwritten methods that may not utilise all the arguments in the signature of the super-class.

## CausalProblem

A CausalProblem can be initialised with just a label, and optionally providing a `Graph` that the instance will store. Said graph will define the relationships between the RVs that the problem considers, and can be set later via the `.graph` attribute if so desired. _Very_ limited functionality (as in, essentially all you can do is set the `.graph` attribute) is available for a `CausalProblem` instance without a set `.graph` attribute.

Otherwise, the primary way a user is expected to interact with a `CausalProblem` is to provide the causal estimand function via the `set_causal_estimand` method. This takes some callable object (`sigma`) which is assumed to be a function of some RVs, with the RVs to be associated to `Node`s in the `.graph`. At present, we also have to run a few "hacks" due to some design aspects we haven't got round to addressing:

- `set_causal_estimand` allows one of the arguments to `sigma` to be a `Graph` object. This is because our current `algorithms` module only has methods that support calls via `expectation(graph, node_name, ...)` etc.
- `sigma` (and the tests for this method) provides a static `rng_key` and `sample_size`, whereas in the long run we probably want these set instance-wise across the whole `CausalProblem` instance.

Once set, the `.causal_estimand` method can be called, which will evaluate the given `sigma` function using the associated `Node`s in `.graph`. This method takes a `jax.Array` as input, which is assumed to be an array of the parameter values - the order of the values matches that of `CausalProblem.graph.parameter_nodes` (hence why it is important for this method to always return parameter nodes in the same order). It then calls `.graph.set_parameters` with the appropriate values extracted from the input array, before running `sigma` and returning the result. This gets around the issue identified in #37 regarding fixing a single realisation of a RV by accident.

`.causal_estimand` primarily exists to be sent into a solver / optimiser - I don't envision the user ever wanting to evaluate it manually, and they certainly should not need to evaluate it manually via a parameter vector. If a user does want to evaluate their causal estimand at a particular set of parameters, the recommended way is to run

```python
# Inputs are sent directly to .graph.set_parameters(**)
causal_problem.set_parameter_values(**parameter_values)
# The .parameter_vector property returns the appropriate vector for the
# current parameter values.
# As such, since we just set the parameter values (using their names),
# we can now get the CP instance itself to build the vector, then evaluate
# the causal estimand at that vector.
causal_problem.causal_estimand(causal_problem.parameter_vector)

### DIFF
--- a/src/causalprog/causal_problem.py
+++ b/src/causalprog/causal_problem.py
@@ -24,7 +24,36 @@ def raises(exception: Exception) -> Callable[[], float]:
 
 
 class CausalProblem(Labelled):
-    """f."""
+    """
+    Container class for handling a causal problem.
+
+    A causal problem <https://github-pages.ucl.ac.uk/causalprog/theory/mathematical-context/>
+    requires an underlying ``Graph`` to describe the relationships between the random
+    variables and parameters, plus a causal estimand and list of (data) constraints.
+    Structural constraints are handled by imposing restrictions on forms of the random
+    variables and constraints directly.
+
+    A ``CausalProblem`` instance brings together these components, providing a container
+    for a causal problem that can be given inputs like empirical data, a solver
+    tolerance, etc, and will provide (estimates of) the bounds for the causal estimand.
+
+    - The ``.graph`` attribute stores the underlying ``Graph`` object.
+    - The ``.causal_estimand`` method evaluates the causal estimand, given values for
+        the parameters.
+    - The ``.constraints`` method evaluates the (vector-valued) constraints, given
+        values for the parameters.
+
+    The user must specify each of the above before a ``CausalProblem`` can be solved.
+    The primary way for this to be done is to construct or load the corresponding
+    ``Graph``, and provide it by setting the ``CausalProblem.graph`` attribute directly.
+    Then, `set_causal_estimand` and `set_constraints` can be used to provide the causal
+    estimand and constraints functions, in terms of the random variables. The
+    ``CausalProblem`` instance will handle turning them into functions of the parameter
+    values under the hood. Initial parameter values (for the purposes of solving) can be
+    provided to the solver method directly or set beforehand via ``set_parameters``. It
+    should never be necessary for the user to interact with, or provide, a vector of
+    parameters (as this is taken care of under the hood).
+    """
 
     _graph: Graph | None
     _sigma: CausalEstimand
@@ -120,7 +149,7 @@ class CausalProblem(Labelled):
         Set the Causal Estimand for this problem.
 
         `sigma` should be a callable object that defines the Causal Estimand of
-        interest, in terms of the random variables of interest to the problem. The
+        interest, in terms of the random variables of to the problem. The
         random variables are in turn represented by `Node`s, with this association being
         recorded in the `rv_to_nodes` dictionary.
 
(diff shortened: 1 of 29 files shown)

### CODE REVIEW

**Comment by mscroggs** [comment]
> The sample method for ParameterNodes [needs to have the same argument names](https://github.com/UCL/causalprog/pull/40/files#diff-3c3e22e486429270877f6243bcb4b4ef3e5c25ae5e72e0391292b25486c23d09) as the super-class for the "up-the-graph" logic to work without an error, so have removed the _ prefixes and slapped on a # noqa for now, until we find a more elegant solution.

I did this exact same change yesterday when I wrote a new test

File: src/causalprog/causal_problem.py [hunk 1]
```diff
@@ -0,0 +1,182 @@
...
+    """Create a callable that raises ``exception`` when called."""
+
+    def _inner() -> float:
+        raise exception
+
+    return _inner
+
+
+class CausalProblem(Labelled):
+    """f."""
```
mscroggs, line 27: This looks like a placeholder docstring

mscroggs (approved):

src/causalprog/causal_problem.py
willGraham01: You know when you really want ruff to stop telling you about something but forget to go back to it? :sweat_smile:

Have filled this in now. Pre-emptively with info about the constraints function too, since I'll be adding that next.

Issue #48, opened 2025-05-06
Remove the `DistributionFamily` class.


willGraham01:
Full context: #47 

## Remove the `DistributionFamily` class

Though we initially thought we would need it, in fact it seems we do not need (and will not need) the functionality the `DistributionFamily` (and subclasses) provide. We can instead just give `DistributionNode._dist` a `Distribution` _class_, and replace statements like `Node._dist.construct(**)` with `Node._dist(**)`.

We should do this just to ensure the changes in the next stages don't have to straddle more classes than necessary.

PR #51, opened 2025-05-07
Make normal sampling faster

mscroggs:
Add `N(m,s) = m + s * N(0, 1)` trick to normal distribution so that samples can be found faster.


mscroggs:
I think it makes sense to remove DistributionFamily (#48) in a separate PR - I first tried to do both together but then figured it made sense to do this first, then do that separately, as this will allow you to experiment more with solvers while I do the other refactoring, as #48 shouldn't now impact the interfaces to distributions

tests/test_causal_problem/test_callables.py
```diff
@@ -94,7 +94,7 @@ def which(request: pytest.FixtureRequest) -> Literal["causal_estimand", "constra
             0.0,
             # Empirical calculation with 1000 samples with fixture RNG key
             # should give 1.8808e-2 as the empirical expectation.
-            2.0e-2,
+            3.0e-2,
```
mscroggs, line 97: These tests started failing, I think due to the rng keys being different in the new way of doing things. Slightly upping the tolerances fixed it. If you don't agree that this is ok, we should look into what I broke

src/causalprog/distribution/family.py
```diff
@@ -67,13 +69,33 @@ def __init__(
...
         return self._member(**parameters)
+
+    def sample(
+        self,
+        samples: int,
+        rng_key: jax.Array,
+        **kwargs: npt.ArrayLike,
+    ) -> npt.NDArray[float]:
+        """Sample values from the distribution."""
+        raise NotImplementedError
```
willGraham01, line 90:
```suggestion
```

Guessing this shouldn't be here (since this is the default "fall-back" method). Plus there's code underneath it which seems to do what we already do as a worst-case option.

src/causalprog/distribution/family.py
```diff
@@ -67,13 +69,33 @@ def __init__(
...
+
+        output = np.zeros(samples)
+        new_key = jax.random.split(rng_key, samples)
+        for sample in range(samples):
+            parameters = {
+                i: j[sample] if hasattr(j, "__len__") else j for i, j in kwargs.items()
+            }
```
willGraham01, line 97:
```suggestion
            parameters = {
                param_name: param_sample[sample] if hasattr(param_sample, "__len__") else param_sample for param_name, param_sample in kwargs.items()
            }
```

Just so it's a bit less obtuse what's going on here now (since we are mixing scalar and array-like samples now).

In the long-term we should probably decide on a more elegant way to handle the scalar/array issue than checking for `__len__`. Though it might be better to test for the `__getitem__` attribute (which allows use of the access `[ ]` syntax on objects) rather than `__len__` as some containers have a length but don't support accessing.

src/causalprog/distribution/normal.py
```diff
@@ -88,3 +88,13 @@ def construct(self, mean: ArrayCompatible, cov: ArrayCompatible) -> Normal:  # t
...
+
+    def sample(
+        self,
+        samples: int,
+        rng_key: JaxArray,
+        **kwargs: npt.ArrayLike,
+    ) -> npt.NDArray[float]:
+        """Sample values from the distribution."""
+        standard = jrn.normal(rng_key, shape=(samples,))
+        return kwargs["mean"] + standard * jnp.sqrt(kwargs["cov"])
```
willGraham01, line 100: Inconvenient fact: `Normal` can represent either a 1D or multivariate normal depending on whether the mean & covariance are given as matrices / vectors.

So in general, the re-parametrisation trick will need to sample as $S = \mu + L *\epsilon$ where $L L^{T}$ is the Cholesky decomposition of the covariance matrix, and $\epsilon$ is an `(samples,)` -shaped vector of draws from $\mathcal{N}(0,1)$.

For scalar (and diagonal) `cov` this is the (element-wise) square-root of course, but for other covariance matrices we'll likely need to do a Cholesky decomposition. Jax does seem to support this at least: <https://docs.jax.dev/en/latest/_autosummary/jax.numpy.linalg.cholesky.html>.

Alternatively, we could just (for the time being) use [`jax.random.multivariate_normal`](https://docs.jax.dev/en/latest/_autosummary/jax.random.multivariate_normal.html) and pass in `kwargs["mean"]` and `kwargs[cov]` (and `shape=(samples,)`)?

src/causalprog/graph/node.py
```diff
@@ -76,20 +76,12 @@ def sample(
...
-            }
-            concrete_dist = self._dist.construct(
-                **parameters, **self._constant_parameters
-            )
-            output[sample] = concrete_dist.sample(new_key[sample], 1)[0][0]
-        return output
+        return self._dist.sample(
+            samples=samples,
+            rng_key=rng_key,
+            **{i: sampled_dependencies[j] for i, j in self._parameters.items()},
```
willGraham01, line 82:
```suggestion
            **{param_name: sampled_dependencies[node_label] for param_name, node_label in self._parameters.items()},
```

tests/test_causal_problem/test_callables.py
willGraham01: Yeah this will do it - we're no longer splitting the key each time we draw a single sample from the distribution (since it's now all done in parallel).

src/causalprog/distribution/family.py
mscroggs: This was some debugging that I forgot to delete

src/causalprog/distribution/family.py
```diff
@@ -67,13 +69,33 @@ def __init__(
...
+        **kwargs: npt.ArrayLike,
+    ) -> npt.NDArray[float]:
+        """Sample values from the distribution."""
+        raise NotImplementedError
+
+        output = np.zeros(samples)
+        new_key = jax.random.split(rng_key, samples)
+        for sample in range(samples):
+            parameters = {
+                i: j[sample] if hasattr(j, "__len__") else j for i, j in kwargs.items()
```
mscroggs:
```suggestion
                i: j[sample] if hasattr(j, "__getitem__") else j for i, j in kwargs.items()
```

src/causalprog/distribution/family.py
```diff
@@ -67,13 +69,32 @@ def __init__(
...
+        rng_key: jax.Array,
+        **kwargs: npt.ArrayLike,
+    ) -> npt.NDArray[float]:
+        """Sample values from the distribution."""
+
+        output = np.zeros(samples)
+        new_key = jax.random.split(rng_key, samples)
+        for sample in range(samples):
+            parameters = {
+                param_name: param_sample[sample] if hasattr(param_sample, "__len__") else param_sample for param_name, param_sample in kwargs.items()
```
mscroggs, line 95:
```suggestion
                param_name: param_sample[sample] if hasattr(param_sample, "__getitem__") else param_sample for param_name, param_sample in kwargs.items()
```

src/causalprog/distribution/normal.py
mscroggs: > Alternatively, we could just (for the time being) use [jax.random.multivariate_normal](https://docs.jax.dev/en/latest/_autosummary/jax.random.multivariate_normal.html) and pass in kwargs["mean"] and kwargs[cov] (and shape=(samples,))?

I tried this and some tests started failing, but they were maybe the same ones I later fixed by adjusting tolerances. I'll try again

willGraham01: If it's just those tests, then it's likely no big deal if the tolerances are still in the same ballpark.

I originally chose the tolerances based on the OoMagnitude that I expected, then stuck a 2 on the front :sweat_smile:

mscroggs:
This is no longer necessary due to being superceeded by numpryo (see #58)

PR #60, opened 2025-08-06
Improve documentation of graph classes

### LINKED ISSUE #53: Documentation of Graph
### Which Section of the Documentation Needs Improving?

Improve the documentation of Graph

### What Can be Improved About This Section

more docs

### How to Improve This Section

_No response_

mscroggs:
- Add args and returns to all docstrings
- Use `@override` so that ruff can see that docstrings are inherits from parent classes - this currently comes from typing_extensions (see #61)
- Make `label` keyword only everywhere

Resolves #53

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,10 @@ lint.ignore = [
     "D212", # multi-line-summary-first-line
     "D407", # removed dashes lines under sections
     "D417", # argument description in docstring (unreliable)
+    "FIX002", # Lines marked with "todo" are OK for now:
     "ISC001", # simplify implicit str concatenation (ruff-format recommended)
+    "TD002", # Lines marked with "todo" are OK for now
+    "TD003", # Lines marked with "todo" are OK for now
 ]
 lint.per-file-ignores = {"__init__.py" = [
     "F401", # Unused import


src/causalprog/graph/graph.py
```diff
@@ -133,6 +200,13 @@ def roots_down_to_outcome(
         Get ordered list of nodes that outcome depends on.
 
         Nodes are ordered so that each node appears after its dependencies.
+
+        Args:
+            outcome_node_label: The label of the outcome node
+
+        Returns:
+            A list of the nodes
```
willGraham01, line 208:
```suggestion
            A list of the nodes, ordered from root nodes to the outcome Node.
```

src/causalprog/graph/node.py
```diff
@@ -38,31 +49,76 @@ def sample(
...
 
     @property
     @abstractmethod
     def constant_parameters(self) -> dict[str, float]:
-        """Named constants that this node depends on."""
+        """
+        Named constants that this node depends on.
+
+        Returns:
+            A dictionary of constant parameters and their values
```
willGraham01, line 108:
```suggestion
            A dictionary of the constant parameter names (keys) and their corresponding values
```

src/causalprog/graph/node.py
```diff
@@ -38,31 +49,76 @@ def sample(
...
 
     @property
     @abstractmethod
     def parameters(self) -> dict[str, str]:
-        """Nodes that this node depends on."""
+        """
+        Nodes that this node depends on.
+
+        Returns:
+            A dictionary of node labels
```
willGraham01, line 119:
```suggestion
        Mapping of distribution parameter names to the nodes that these are represented by.

        Returns:
            Mapping of distribution parameters (keys) to the corresponding label of the node that represents this parameter (value).
```

willGraham01 (approved):
Made a few suggestions for some clarifications.

Also gave a suggestion for the "What is a Parameter" comment that has been left in (and the corresponding docstring for Node.parameters). Feel free to iterate on this.

src/causalprog/graph/graph.py
```diff
@@ -41,11 +63,16 @@ def add_node(self, node: Node) -> None:
...
+        Add a directed edge to the graph.
 
         Adding an edge between nodes not currently in the graph,
         will cause said nodes to be added to the graph along with
         the edge.
+
+        Args:
+            first_node: The node that the edge points from
+            second_node: The node that the edge points to
+
```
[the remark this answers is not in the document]
mscroggs: "leaves" feels dangerous as leaf nodes are something else. How about "start_node" and "end_node"?

Issue #62, opened 2025-08-06
Organising the codebase


willGraham01:
### Is Your Feature Request Related to a Problem? Please Describe

Currently our package codebase and test suite is a bit all over the place. The test suite has a lot of repeated setup that we would ideally move into fixtures, or into setup fixtures.

The package code itself has a number of files that contain multiple "important" classes that we should probably split out for our own sanity down the line.

TLDR; It is probably time to introduce some order.

## Test Suite

The test suite is the easy one to tackle, in which I propose we organise the test suite along the lines of:

```default
tests/
    fixtures/
    test_unit/
    test_integration/
    conftest.py
```

We can setup `conftest.py` to automatically search the `fixtures` directory (and any sub-directories) for `pytest.fixture` objects, which all tests will have access to. We can also put any "utility fixtures" in here, like helpers for `pytest.raise` contexts.

`test_unit` can then follow the structure that we seem to largely be adhering to currently. Small unit tests for methods should go in here, or inside sub-folders appropriately. Personally I prefer one-test-file-per-method-being-tested, rather than one-file-per-class-or-module, doubly so since our test files already seem to be quite long.

`test_integration` may or may not be needed, but essentially serves as a space for some of our larger examples. Like when we eventually introduce #45 ?

## Codebase Itself

Here, I think we just need to set some rules for how modular we are. Nothing is immediately pressing, though we could  be a bit more modular with our files and where we define classes. The `node.py` file for example defines three classes, for example - these should probably be 3 separate files within their own sub(sub)module.

mscroggs:
Test suite is now modularised. Discussion still needed on modularity of the main codebase

PR #67, opened 2025-08-07
Organise the test suite

willGraham01:
Concerns #62 |

Implements the structure for the test suite as set out in #62. The test suite now looks like:

```default
tests/
    fixtures/
    <unit tests, broken down into subfolders>/
    test_integration/
    conftext.py
```

`conftext.py` is setup to automatically discover all fixtures defined inside the `fixtures` directory when the tests are run, which means we can reduce code duplication across the test suite where possible. In particular, this means that the TODOs that were present in #59 have now been addressed.

I haven't updated or moved around anything else in the test suite, so we'll need to go through our other tests to check if they conform to this new structure at some point (again, #62). I also didn't introduce the `test_unit` subfolder, but we can do this if we want to (though it will involve moving a whole bunch of files inside `tests/` to `tests/test_unit`, making the PR noisy).

Also updated the docs build to include this information, and point to a few "useful" fixtures that will likely have widespread use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![Tests status][tests-badge]][tests-link]
 [![Linting status][linting-badge]][linting-link]
 [![Documentation status][documentation-badge]][documentation-link]
-[![License][license-badge]](./LICENSE.md)
+[![License][license-badge]](http://github-pages.ucl.ac.uk/causalprog/LICENSE.md)
 
 <!-- prettier-ignore-start -->
 [tests-badge]:              https://github.com/UCL/causalprog/actions/workflows/tests.yml/badge.svg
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -4,7 +4,7 @@
 [![Tests status][tests-badge]][tests-link]
 [![Linting status][linting-badge]][linting-link]
 [![Documentation status][documentation-badge]][documentation-link]
-[![License][license-badge]](./LICENSE.md)
+[![License][license-badge]](http://github-pages.ucl.ac.uk/causalprog/LICENSE.md)
```
mscroggs, line 7:
```suggestion
[![License][license-badge]](http://github-pages.ucl.ac.uk/causalprog/LICENSE)
```

mscroggs (approved):

PR #69, opened 2025-08-08
Tidy up graph tests

mscroggs:
- Use fixtures for graphs whenever used by multiple tests
- Move tests into test_graph folder and split up out of the long test file into three files
- Use [factories as fixtures](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#factories-as-fixtures)
- Update tests and DistributionNode.sample to use numpyro: Fixes #25. Fixes #26. This will make #48 easier to do.

See #62. See #58.

Issue #58, opened 2025-08-05
NumPyro for Sampling


willGraham01:
### Is Your Feature Request Related to a Problem? Please Describe

We should investigate `NumPyro` as a possible library for handling the (Python object that describes) the distributions, and samples from them.

### Describe the Solution You'd Like

`Node` (or `DistributionNode`) should have a method of constructing the "random variable" (or more precisely, the distribution describing that RV) from the information stored in the `Node`.

For now, we will assume we're only using `NumPyro` distributions, and we may re-introduce the backend-agnosticity later.

### Additional Context

![Image](https://github.com/user-attachments/assets/da5bb3dd-5a0e-42d7-b29a-12ce592e6fe1)

willGraham01:
#59 now lets us create "parameterised" models from our `Graph`. That is to say, one can call

```python
constructor = graph.model_constructor()
```

The `constructor` obtained is itself another function, which is called using

```python
realisation = constructor(**parameter_node_values)
```

and will produce `realisation`, which is the graph model __with the parameter nodes set to the `param_node_values`__, coded up in a way that `numpyro` understands. `realisation` can thus be used can be used as a sampler for MCMC algorithms, for example:

```python
constructor = graph.model_constructor()
realisation = constructor(**parameter_node_values)

kernel = numpyro.infer.NUTS(realisation)
mcmc = numpyro.mcmc(kernel)
mcmc.run()

# Samples can then be fetched using
mcmc.get_samples()[node_label]
```

This means that we do now have a way to draw samples from a model, if we have been told the parameter node values beforehand. In particular, we can now empirically estimate the moments of RVs using our newfound ability to generate appropriate samples.

This is still not ideal for our needs when looking at the optimisation problem. We have to run an MCMC sampler _every time_ we want to evaluate a particular moment of a random variable, since we have to re-build the `realisation` each time the parameter nodes change value.

PR #70, opened 2025-08-08
Add moment function

mscroggs:
- Add a moment function that estimates the moment of any order from a graph.

Follows on from #69, so we should merge that first.


src/causalprog/algorithms/moments.py
```diff
@@ -43,4 +43,22 @@ def standard_deviation(
     rng_key: jax.Array,
 ) -> float:
     """Estimate the standard deviation of a graph."""
-    return sample(graph, outcome_node_label, samples, rng_key=rng_key).std()
+    return (
+        moment(2, graph, outcome_node_label, samples, rng_key=rng_key)
+        - moment(1, graph, outcome_node_label, samples, rng_key=rng_key) ** 2
```
mscroggs, line 48: I'm unsure here if we want to split the rng_key (and so use different samples for the two moments) or use the same key so that the two moments use the same sample data

src/causalprog/algorithms/moments.py
```diff
@@ -32,7 +32,7 @@ def expectation(
     rng_key: jax.Array,
 ) -> float:
     """Estimate the expectation of a graph."""
-    return sample(graph, outcome_node_label, samples, rng_key=rng_key).mean()
+    return moment(1, graph, outcome_node_label, samples, rng_key=rng_key)
```
willGraham01, line 35: I can't comment on the line above so leaving a comment nearby, but all the docstrings here refer to the graph as the object we are taking moments/samples/etc from. But in actual fact it's the (RVs attached to the) nodes right?

So 

```python
"""Estimate the expectation of a graph."""
```

should probably be

```python
"""Estimate the expectation of (the RV attached to) a node in the graph."""
```

Similarly for the other functions here.

src/causalprog/graph/node.py
```diff
@@ -178,19 +178,20 @@ def sample(
         samples: int,
         rng_key: jax.Array,
     ) -> npt.NDArray[float]:
+        d = self._dist(
```
willGraham01, line 172: Again this is more a note rather than something that needs to be done in this PR; but now that we're sticking to `numpyro` we can avoid implementing this sequential sampling down the nodes we've got going right now, and instead use `Graph.model` to generate the samples for us.

src/causalprog/graph/node.py
```diff
@@ -178,19 +178,20 @@ def sample(
...
-                    native_name: sampled_dependencies[node_name]
-                    for native_name, node_name in self.parameters.items()
-                },
-                # Pass in any constant parameters this node sets
-                **self.constant_parameters,
-            ),
+            d,
             rng_key=rng_key,
-            sample_shape=(samples,),
+            sample_shape=(samples,) if d.batch_shape == () and samples > 1 else (),
```
willGraham01, line 185: Likewise, see my comment above. Using the `Graph.model` will save us having to do things like this.

willGraham01 (approved):
The `moment` and other functions implemented look good, so this is OK to merge. Though one thing that catches my attention is that we can now use `numpyro` to draw the samples from the `Graph.model`, rather than doing our current "sample down the tree" implementation. This should also fix our current issues with test speed too, but I'll open a separate issue for this.

That, and (from my understanding) if a `ParameterNode` does not have a current value set, the `sample` and thus other moment methods will throw an error if we try to use them? But this is a problem I think we can sidestep by switching to using `Graph.model` in the `sample` method instead.

Approving the PR because none of these problems are introduced in this PR (they're already in `main` since we're currently meshing numpyro into everything). We can fix these in another PR.

willGraham01:
(Disregard the above comment, I forgot that we already discussed this in https://github.com/UCL/causalprog/issues/58#issuecomment-3178653891 ). We're already planning to swap to doing an MCMC using `Graph.model` )

src/causalprog/algorithms/moments.py
mscroggs: This seems like a sensible solution that we can revisit later if needed

PR #71, opened 2025-08-11
Remove complexity around `Graph.model_constructor`

willGraham01:
Removes the `Graph.model_constructor` method and replaces it with `Graph.model`. Amongst other things:

- This is more in-line with how `numpyro` (and the wider stochastic programming community) seem to expect models to be defined.
- It saves us a couple of layers of nested / wrapped functions.

`Graph.model` is effectively the "function" that was wrapped and returned by `Graph.model_constructor`, however there is actually no need to do this. Instead, the "model function" can simply be defined to take the parameter values as input arguments, and then these parameter values are passed into whichever function/method we want to use on said method, like MCMC.

Compare before:

```python
g = Graph(...)

constructor = g.model_constructor()
realisation = constructor(**parameter_values)

kernel = numpyro.infer.NUTS(realisation)
mcmc = numpyro.infer.MCMC(kernel)
mcmc.run(rng_key)
```

to after

```python
g = Graph(...)

kernel = numpyro.infer.NUTS(g.model)
mcmc = numpyro.infer.MCMC(kernel)
mcmc.run(rng_key, **parameter_values)
```

Notably, we can setup the MCMC once, and then call it on the model multiple times with different parameter values. In contrast, before we would have to create one MCMC instance per set of parameter values we wanted to use.

---

Shouldn't cause any conflicts, unless `node.sample` or `graph.sample` have been updated in another PR to use the `Graph.model_constructor` method.

### DIFF
--- a/src/causalprog/graph/graph.py
+++ b/src/causalprog/graph/graph.py
@@ -219,25 +219,7 @@ class Graph(Labelled):
 
         The method returns a dictionary recording the mode sites that are created.
         This means that the model can be 'extended' further by defining additional
-        sites in a wrapper around this method. For example,
-        >>> mu_x = ParameterNode(label="mu_x")
-        >>> x = DistributionNode(
-                numpyro.distributions.Normal,
-                label="X",
-                parameters={"loc": "mu_x"},
-                constant_parameters={"scale": 1.0},
-            )
-        >>> g = Graph(label="One normal")
-        >>> g.add_edge(mu_x, x)
-        >>> def extended_model(*, nu_y, **parameter_values):
-        ...     sites = g.model(**parameter_values)
-        ...     numpyro.sample(
-        ...         "Y",
-        ...         numpyro.distributions.Normal(
-        ...             loc=sites["X"],
-        ...             scale=nu_y,
-        ...             ),
-        ...         )
+        sites in a wrapper around this method.
 
         Args:
             parameter_values: Names of the keyword arguments should match the labels


src/causalprog/graph/graph.py
```diff
@@ -211,43 +209,59 @@ def roots_down_to_outcome(
...
-        or in this implementation the values of the `ParameterNode`s. This means that
-        a model can only be specified (and then used to sample from, for example) once
-        values for the `ParameterNode`s have been given. However, each of these models
-        is built in the same way, and all that needs to be done is substitute the new
-        parameter values for the old.
+    def model(self, **parameter_values: npt.ArrayLike) -> dict[str, npt.ArrayLike]:
         """
+        Model corresponding to the `Graph`'s structure.
+
+        The model created takes values of the `ParameterNode`s as keyword arguments.
```
mscroggs, line 216: `self.parameter_nodes` will include all nodes where `is_parameter` is `True` - currently this is only `ParameterNode`s, but in future this could be something else. Maybe updating the docs something like this is sensible?

```suggestion
        The model created takes values of the nodes that are parameter as keyword arguments.
```

tests/test_graph/test_model.py
```diff
@@ -0,0 +1,145 @@
...
+def two_normal_graph_expected_model() -> Callable[..., dict[str, npt.ArrayLike]]:
+    """Creates the model that the two_normal_graph should produce."""
+
+    def _inner(mu_x: float, nu_y: float) -> dict[str, npt.ArrayLike]:
+        x = numpyro.sample("X", Normal(loc=mu_x, scale=1.0))
+        y = numpyro.sample("Y", Normal(loc=x, scale=nu_y))
+
+        return {"X": x, "Y": y}
+
+    return _inner
```
mscroggs, line 48: We should move these to fixtures when we merge #69

tests/test_graph/test_model.py
```diff
@@ -0,0 +1,145 @@
...
+    g.add_node(nu_y)
+    g.add_edge(mu_x, x)
+    g.add_edge(nu_y, y)
+    g.add_edge(x, y)
+
+    return g
+
+
+@pytest.fixture
+def two_normal_graph_expected_model() -> Callable[..., dict[str, npt.ArrayLike]]:
```
mscroggs, line 39: This looks to me to be the same as the `ux_x_graph` I added in #69

mscroggs (approved):
If we merge this first, I can then update #69 to deal with the comments I left about test tidying.

Issue #75, opened 2025-08-13
Solving the Optimisation Problem


willGraham01:
### Is Your Feature Request Related to a Problem? Please Describe

Related to #38 

We are going to need a jax-friendly way of solving the minimisation problem

$$ \mathrm{min}_{\theta} f(\theta), \quad\text{s.t. } g(\theta) \leq 0. $$

### Describe the Solution You'd Like

A method (either written ourselves or from an external library) that can solve the above problem. We'd probably need to use [the KKT algorithm](https://en.wikipedia.org/wiki/Karush%E2%80%93Kuhn%E2%80%93Tucker_conditions#Generalizations) (the generalisation of the method of Lagrange multipliers) but can make the small simplification that there are _no_ equality constraints.

We could do this manually ourselves, following a _very_ naive method, that ignores the stochastic element of our $f$ and $g$ functions:

1. Assemble the augmented Lagrangian

$$ \mathcal{L}(\theta, \lambda) := f(\theta) + \lambda g(\theta), $$

as a callable, jax-friendly Python function. (This should be easy enough if we have a way of evaluating $f$ and $g$ already).

2. Set $\mathbf{x} = (\theta, \lambda)$ and $\mathbf{F}(\mathbf{x}) := \nabla\mathcal{L}(\mathbf{x})$, again in a jax-friendly way.

3. We solve $\mathbf{F}(\mathbf{x}) = 0$. This will require something like Newton-Raphson; where we take an initial guess $\mathbf{x}_0$ and iteratively solve

$$ \mathbf{F}(\mathbf{x}_n) = \nabla\mathbf{F}(\mathbf{x}_n) \mathbf{y} $$

for $\mathbf{y}$, then update $\mathbf{x}_{n+1} = \mathbf{y} + \mathbf{x}_n$.

4. Given the solution $\mathbf{x}\_{sol} = (\theta_{sol}, \lambda_{sol})$ of part (3), we then check that $\lambda_{sol} \geq 0$. If this is true, then $\theta_{sol}$ is a (local) minimiser.

This has a number of problems;

- The stochastic manner in which we evaluate $f$ and $g$ is ignored.
- It requires an initial guess of the suspected optimal parameters, but also the Lagrange multiplier $\lambda$.
- It requires computing the Hessian of $\mathcal{L}$!

### Describe Alternatives You've Considered

[...]

PR #76, opened 2025-08-13
Integration test for simple example

### LINKED ISSUE #38: Create an integration test for the simple example
The simple example in #34 should have an integration test written for it. Even if it doesn't work right away, we need it to test UI and compatibility of features.

We take two random variables $X$ and $Y$ with the following structural equations;

$$ f_X = \mathcal{N}(\mu_X, \nu_X^2), \quad f_Y = \mathcal{N}(X, \nu_Y^2), $$

where we will assume, for the time being, that $\nu_X, \nu_Y \in [0,\infty)$ are fixed values, and $\mu_X$ is a parameter in our causal model.
This means that the causal model, $\Theta = \{\mu_X\}$ has only one parameter for the time being.

Our causal estimand of interest will simply be the expectation of our "outcome" variable $Y$;

$$ \sigma = \mathbb{E}[Y]. $$

Our constraints will be the observed mean of $X$, $\phi = \mathbb{E}[X]$, and we will take our distance function to be $\mathrm{dist}(\phi, \psi) = \vert \phi - \psi \vert$ (essentially a 1-dimensional $L^2$-norm).
For a given tolerance $\epsilon$, we thus have the following problem to solve;

$$ \max_{\mu_X} / \min_{\mu_X} \mathbb{E}[Y], \quad\text{subject to}\quad \vert \phi_{\mathrm{data}} - \mathbb{E}[X] \vert \leq \epsilon. $$

By the structural equations, we can infer that $\mathbb{E}[Y] = \mathbb{E}[X] = \mu_X$, thus reaching

$$ \max_{\mu_X} / \min_{\mu_X} \mu_X, \quad\text{subject to}\quad \vert \phi_{\mathrm{data}} - \mu_X \vert \leq \epsilon. $$

The solution to this problem is $\mu_X = \phi_\mathrm{data} \pm \epsilon$ (the positive solution corresponding to the maximisation).

willGraham01:
Resolves #38, though #75 remains something that we need to address.

Implements a barebones integration test for the simple example problem involving two normals.

All the steps to setup the causal estimand and constraints are done manually, IE by manipulating the `Graph.model` rather than attempting to use the `CausalProblem` class. Iterative PRs on top of this one will slowly take more of the manual steps out of the new integration test and into the appropriate classes.

Also, the integration test currently uses the "minimise" approach described in #75, whereby to find stationary points of the Lagrangian we elect to instead use SGD to minimise $\vert\nabla\mathcal{L}\vert^2$. This _works_ for this particular example, but we also have the benefit of knowing the exact solution so are able to provide a very good initial value.


willGraham01:
https://github.com/UCL/causalprog/pull/70 changes the names of the nodes + parameters inside the `two_normal_graph_parametrized_mean` fixture, will need to update text in light of this PR if it goes in first.

Issue #27, opened 2025-03-20
Standardise type hints


mscroggs:
We should standarise type hints for things like "array-like" and "rng seed"

willGraham01:
I think our plan for this, for now, should be to make a `_typing` module or something, and inside it create a few `TypeVar`s or `TypeAlias`es that cover the things we need to standardise.

Then go through the package and typehint accordingly - that should hopefully avoid us having to do this again some time in the future... It also lets us helpfully avoid commitment to a particular framework for now (though `jax` should probably be our default).

EG (imports / correct class names not promised)

_typing.py
```python

RNGKey: TypeAlias = jax.Array | numpy.Generator
ArrayLike: TypeAlias = jax.Array | numpy.typing.ArrayLike
```

some_other_file.py
```python
from causalprog._typing import RNGKey


def random_function(rng: RNGKey, ...):
    ...
```

PR #77, opened 2025-08-13
Remove DistributionFamily

### LINKED ISSUE #48: Remove the `DistributionFamily` class.

mscroggs:
Resolves #48.

We may want to revert the second commit (https://github.com/UCL/causalprog/commit/e8006911f2a8d1b722e18ab771aae829bbf96452) if we want to keep the base of distribution?


src/causalprog/graph/node.py
```diff
@@ -143,7 +141,7 @@ class DistributionNode(Node):
 
     def __init__(
         self,
-        distribution: DistributionFamily,
+        distribution: type,
```
willGraham01, line 144: There's probably a more helpful typehint we can put here (with enough effort), but that can be a bonus job for https://github.com/UCL/causalprog/issues/27

willGraham01 (approved):
> We may want to revert the second commit (https://github.com/UCL/causalprog/commit/e8006911f2a8d1b722e18ab771aae829bbf96452) if we want to keep the base of distribution?

We can always do this in the future if we decide we want it back. Right now we're not using it given we've committed to `numpyro` for the time being, so I think it's safe to let it go here.

When / if we do come back to it, we can always fetch this from the ether again.

PR #78, opened 2025-08-14
Remove value from parameter nodes

### LINKED ISSUE #72: Remove value from ParameterNode
Following #58, we'll no longer need to be able to set a value for a ParameterNode

mscroggs:
- Removes `value` from `ParameterNode`: resolves #72
- Removes `set_parameters` from `Graph`
- Adds `parameter_values` input to functions where it will now be needed
- Adds PLR0913 (Function has too many arguments) to skips (See #19)


tests/test_causal_problem/test_callables.py
```diff
@@ -10,47 +10,72 @@
 from causalprog.causal_problem import CausalProblem
 from causalprog.graph import Graph, Node
 
-
-@pytest.fixture
-def n_samples_for_estimands() -> int:
-    return 1000
+n_samples_for_estimands = 1000
```
willGraham01, line 13:
```suggestion
@pytest.fixture
def n_samples_for_estimands() -> int:
 return 1000
```

We shouldn't use a module-scoped variable for this, in the off chance that it gets accidently reassigned during one of the tests.

If you want to avoid a fixture, you can explicitly define it as an argument that takes a constant value to the tests / other fixtures that require it. EG

```python
def expectation_fixture(
    rng_key: jax.Array,
    n_samples_for_estimands: int = 1000,
)
```

tests/test_causal_problem/test_graph_and_param.py
```diff
@@ -7,7 +8,9 @@
 from causalprog.graph import Graph
 
 
-def test_graph_and_parameter_interactions(graph: Graph) -> None:
+def test_graph_and_parameter_interactions(
+    two_normal_graph: Callable[..., Graph],
+) -> None:
     cp = CausalProblem(label="TestCP")
 
     # Without a graph, we can't do anything
```
willGraham01, line 16: Can't comment on non-diff lines again, but I noticed that we can use the `raises_error` fixture here instead of the explicit `pytest.raises` block.

tests/test_graph/test_parameters.py
```diff
@@ -5,83 +5,20 @@
...
-
-
 def test_parameter_node(rng_key):
     node = ParameterNode(label="mu")
 
-    with pytest.raises(ValueError, match="Cannot sample"):
-        node.sample({}, 1, rng_key)
-
-    node.value = 0.3
+    with pytest.raises(ValueError, match="Missing input for parameter"):
```
willGraham01, line 19:
```suggestion
    with raises_context(ValueError("Missing input for parameter")):
```

(will also need to use the `raises_context` fixture)

src/causalprog/algorithms/moments.py
```diff
@@ -20,7 +21,12 @@ def sample(
     keys = jax.random.split(rng_key, len(nodes))
 
     for node, key in zip(nodes, keys, strict=False):
-        values[node.label] = node.sample(values, samples, rng_key=key)
+        values[node.label] = node.sample(
+            {} if parameter_values is None else parameter_values,
```
willGraham01, line 25:
```suggestion
            parameter_values if parameter_values else {},
```

Depending on how you prefer your Python syntax-tic sugar

willGraham01 (approved):

tests/test_causal_problem/test_callables.py
mscroggs: I changed this while trying to work out why a test was failing and probably shouldn't have committed this bit. Will revert

PR #101, opened 2025-09-24
Use objects for solver outcomes, not errors and tuples

willGraham01:
TLDR;

- Creates a `SolverResult` class that can be used as the output for any methods we implement to solve casual problems.
- Replaces the return type of the SGD solver method with this class, populating appropriate values.

## Why?

A couple of reasons.

The first is that the SGD method was already returning a tuple of 4 values, the order of which was both somewhat arbitrary and going to be a pain to enforce consistently across the codebase if we ever add any more solver methods. Having a simple container object that can store this information for later access is cleaner to return.

The second being that we don't want to rely on throwing errors when one of our solver methods fails to converge (or encounters some other problem), since we may want to take action based on this result. (This is particularly relevant to the Quadratic Penalty Method). Instead, we should just bundle all the relevant information into the object that's returned, which can indicate whether or not the solver method was successful on inspection of an attribute.

### DIFF
--- a/src/causalprog/solvers/sgd.py
+++ b/src/causalprog/solvers/sgd.py
@@ -110,7 +110,7 @@ def stochastic_gradient_descent(
     )
 
     return SolverResult(
-        arg_result=current_params,
+        fn_args=current_params,
         grad_val=gradient_value,
         iters=iters_used,
         maxiter=maxiter,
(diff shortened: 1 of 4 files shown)


samjmolyneux (approved):
Looks good to me!

PR #116, opened 2025-10-30
Sjmolyneux/two std normal notebook

### LINKED ISSUE #103: Turn two normal example into a notebook
### Is Your Feature Request Related to a Problem? Please Describe

We should turn the test in `test_integration/two_normal_example.py` into a notebook that can be used as an interactive introduction to the package.

### Describe the Solution You'd Like

For the time being, convert the steps in `test_integration/two_normal_example.py` into a Jupyter Notebook that (in a suitable environment with `causalprog` installed) runs to completion, with markdown cells between key steps to explain what is happening.

For completeness's sake, the notebook should also outline / define the problem that is being setup and solved (mathematically) along with explaining the code that is being used to do this.

If committing the notebook to the repository, it should probably live in a top-level `examples/` folder.

### Describe Alternatives You've Considered

_No response_

### Additional Context

A possible extension to this is, once we have such a notebook, we can use a suitable tool to have it tracked and run by the CI, and included as part of the documentation build.

samjmolyneux:
1. Added a test that runs all notebooks in the examples folder to ensure that they run without error. This also included the addition of [nbmake](https://github.com/treebeardtech/nbmake) to the test dependencies. 

2. We agreed not to use [Jupytext](https://github.com/mwouts/jupytext) for the time being, so I added an [nbstripout](https://github.com/kynan/nbstripout) hook to the pre-commit. This refuses to commit notebooks until all meta data has been stripped. 

3. Of course 1. and 2. can be removed if we later decide to use [Jupytext](https://github.com/mwouts/jupytext).

4. Added `examples/two_normal.ipynb`. Notebook builds up an example graph using two normal distributions. Notebook then attempts to solve by running sgd on $\lVert \nabla \mathcal{L} \rVert_2^2$. Notebook then demonstrates that the objective function can be easily changed and solves the problem using a custom objective function based up minimising KKT residuals. 

5. Added some ruff ignores rules to examples folder. I think we probably don't care about these things missing in examples.  

Closes #103.


tests/test_notebooks/test_notebooks_run.py
```diff
@@ -0,0 +1,89 @@
...
+    expect_test_file = (
+        root_directory / "tests" / "test_notebooks" / "test_notebooks_run.py"
+    )
+    assert (
+        expect_test_file.is_file()
+    ), """The structure of the tests have changed, this means that this test will fail
+         to find the notebooks in this repo. If this test file has moved, update the
+         root_directory definition and this assertion accordingly."""
+
+    assert notebook_paths, f"No notebooks found under {examples_directory}"
```
willGraham01: Whilst it's true that we do depend on this setup, I don't think we need to actively have a test for this. When writing tests, the project structure is assumed to be setup correctly.

If we change this structure without catching something here, our tests will fail anyway for precisely the same reasons as this test flags. So having an explicit test for this isn't necessarily adding value to the test suite.

tests/test_notebooks/test_notebooks_run.py
```diff
@@ -0,0 +1,89 @@
...
+
+root_directory = Path(__file__).resolve().parents[2]
+examples_directory = root_directory / "examples"
+
+# Hidden directories are excluded to avoid testing checkpoints
+notebook_paths = [
+    str(path)
+    for path in examples_directory.rglob("*.ipynb")
+    if not any(part.startswith(".") for part in path.parts)
+]
```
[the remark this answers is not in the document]
willGraham01: Yeah looks like it's either functions, or a hook into `pytest_generate_tests` which is not a rabbit hole I want to explore.

What you suggested looks good. Note that `test_notebooks_run` will need the `root_directory` variable, but I think `root_directory` does deserve to be promoted to a fixture in it's own right (and doing so shouldn't hit the same parameter issues);

`tests/fixtures/general.py`
```python
@pytest.fixture
def root_directory() -> Path
    """Absolute path to the repository root."""
    return Path(__file__).parents[2].resolve()
```

tests/test_notebooks/test_notebooks_run.py
samjmolyneux: But won't this test pass no matter what? 
We are getting notebook_paths by running `Path.rglob` on the root. Then we are getting `n_discoverable_notebooks` by running `glob.glob` on the root. So we are testing if`Path.rglob` and `glob.glob` are the same. 

What we really want to know is that all of our notebooks are being found, one way would be to keep a list of the files or a counter and test that they have all been found. Or we could test that the root has been found correctly. 

We could use 


And again, if the fixture were to move, albeit far less likely in it's new location, the test would become redundant whilst still always passing.

samjmolyneux: > For non-developer and non-editable (pip -e) installs, as_file(files(causalprog)) will direct us to the install location of the environment's Python executable, which won't have the notebooks present (since examples/ is not packaged with causalprog).

Yeah I suppose I didn't consider the fact that someone may run the tests in the repo without having the package installed in editable mode. 

Will make the changes now.

willGraham01 (approved):

Issue #129, opened 2026-06-04
Implement r builder


mscroggs:
For now, hard code that we're using the graph from #126. (see also #130)

- [ ] `r_builder`
   - Inputs: graph, values that are now fixed (ie learnt data at $\theta_x$), QuadratureMethod (#128)
   - Returns: callable `r(theta, x, z, l) -> float | Vec<f64>`
   - This r function is Ricardo's r function. To evaluate this, we need to:
       - compute $v_y$ and $m_y$
       - get points and weights from QuadratureMethod
       - compute $f_y(...)$ and $\pi_{ul}(...)$
       - compute $r(\theta)$ via sum and return

willGraham01:
Done in #147

PR #140, opened 2026-06-04
Add quadrature classes

### LINKED ISSUE #128: Implement a QuadratureMethod class
- [ ] Make an ABC:
    - Class methods:
        - `__init__(npoints)
        - `@property npoints() -> int`
        - `@abstractmethod points_and_weights() -> tuple[NDArray, NDArray]`
- [ ] Make concrete subclasses:
    - [ ] GaussianQuadrature:
        - Inputs: npoints. Make Gauss quadrature rule
    - [ ] MonteCarloGaussian:
    - Inputs: n points.
    - `points_and_weights` returns (random samples from std Gaussian, `vec![1.0 / n as f64; npoints]`)

mscroggs:
Close #128

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,16 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
     "Typing :: Typed",
 ]
-dependencies = ["jax", "networkx", "numpy", "numpyro", "optax", "tqdm", "typing_extensions"]
+dependencies = [
+    "jax",
+    "networkx",
+    "numpy",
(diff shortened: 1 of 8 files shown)


willGraham01 (approved):
Approving as this was something that largely came out of our prep-day too.

Having started working on #129, it would be useful for the `QuadratureMethod` class to enforce the definition of an `integrate` method too, since this logic will depend on the interval width / sample points etc that is specific to the method chosen. But this can come in a later PR.

Issue #124, opened 2026-06-04
RVNodes


mscroggs:
- [x] Add ContinuousRandomVariableNode and DiscreteRandomVariableNode
- [x] See also information about `evaluate` in #123. This should:
   - For the node C, the value it currently takes should be input into evaluate, then evaluate return this.
   - For the nodes UX, UY, X, Y, it returns f_*(...). Note that this is deterministic if the nodes this depends on have been determined
   - `panic!` if any parent nodes have no determined value

PR #141, opened 2026-06-05
rename ParameterNode to DataNode

### LINKED ISSUE #123: DataNode
- [x] Rename ParameterNode to DataNode
- [x] Add `evaluate` to BaseNode class. This should evaluate the nodes given all their inputs and `panic!` if any input is missing or if node is DistributionNode

mscroggs:
- Rename node
- Add `evaluate` method

Resolves #123


src/causalprog/graph/node/base.py
```diff
@@ -127,6 +127,22 @@ def sample(
...
+        **given_values: dict[str, float | npt.NDArray[float]],
+    ) -> float | npt.NDArray[float]:
+        """
+        Evaluate the node.
+
+        Args:
+            given_values: Values for data nodes and values of parents
+
+        Returns:
+            Value of this node given the given values
```
willGraham01:
```suggestion
            Value of this node given `given_values`.
```

src/causalprog/graph/node/component.py
```diff
@@ -48,6 +49,17 @@ def sample(
...
 
+    @override
+    def evaluate(
+        self,
+        **given_values: dict[str, float | npt.NDArray[float]],
+    ) -> float | npt.NDArray[float]:
+        parent_value = given_values[self._parent_node_label]
+        if not isinstance(parent_value, np.ndarray):
+            msg = f"Invalid data in node: {self._parent_node_label}"
+            raise TypeError(msg)
```
willGraham01: This is likely going to cause trouble since we could have a mix of `jax`, `numpy` and potentially `torch.Tensor`s in here.

Maybe we do the python thing and "try and beg forgiveness later". Namely don't bother with this check at all, since Python isn't strongly typed anyway.

The other alternative is to check against all variable types that we reasonably expect here (`jax.numpy.Array`, `torch.Tensor`, etc), but that doesn't seem particularly scalable.

src/causalprog/graph/node/data.py
```diff
@@ -44,17 +38,35 @@ def sample(
...
+            msg = f"Missing input for node: {self.label}."
+            raise ValueError(msg)
+        value = given_values[self.label]
+        if self.shape == ():
+            if not isinstance(value, float):
+                msg = f"Invalid value for note: {self.label}"
+                raise ValueError(msg)
+        elif not isinstance(value, np.ndarray) or self.shape != value.shape:
+            msg = f"Invalid value for nose: {self.label}"
+            raise ValueError(msg)
```
willGraham01:
```suggestion
        elif self.shape != value.shape:
            msg = f"Invalid value for nose: {self.label}"
            raise ValueError(msg)
```

Again, we might not have only `numpy` arrays floating around here. I think we just remove the hard restriction, and accept that we'll get `TypeErrors` if we try and do silly things.

tests/test_algorithms/test_evaluate.py
```diff
@@ -0,0 +1,13 @@
...
+
+from causalprog.graph import DataNode
+
+
+def test_evaluate_node(raises_context):
+    node = DataNode(label="A")
+    assert np.isclose(node.evaluate(A=2.0), 2.0)
+
+    with raises_context(ValueError("Missing input for node: A")):
+        node.evaluate()
```
willGraham01: It's probably worth being granular, as when we introduce `ContinuousRVNode` and `DiscreteRVNode` via #124, we'll have a lot more cases to check via `evaluate`.

Also we should probably check that `DistributionNode.evaluate()` is behaving correctly here (which I've added in my suggestion below, check I'm right). Similarly, we probably need some checks for what `ComponentNode.evaluate()` is doing too? Suggestions below, along with parametrisation ideas, feel free to edit as necessary (and the linter I'm sure with have thoughts on this too).

```suggestion
@pytest.mark.parametrize(
    ("node", "kwargs_to_evaluate", "expected_result"),
    [
        pytest.param(DataNode(label="A"), {"A": 2.0}, 2.0, id="Evaluate DataNode itself"),
        # Likely want a test here for a ComponentNode?
        pytest.param(ComponentNode("Parent", 1, label="Child"), {"Parent": np.arange(4)}, 1.0, id="Evaluate ComponentNode, given parent"),
    ]
)
def test_evaluate_node(node, kwargs_to_evaluate, expected_result):
    assert np.allclose(node.evaluate(**kwargs_to_evaluate), expected_result)

@pytest.mark.parametrize(
    ("node", "kwargs_for_evaluate", "expected_error"),
    [
        pytest.param(DataNode(label="A"), {}, ValueError("Missing input for node: A"), id="DataNode missing input value"),
        pytest.param(DistributionNode(dist=None, label="A"), {}, ValueError("Cannot evaluate a DistributionNode"), id="Attempt to evaluate DistributionNode"),
    ]
)
def test_evaluate_node_fail_on_missing_data(node, kwargs_for_evaluate, expected_error, raises_context):
    with raises_context(expected_error):
        node.evaluate(**kwargs_for_evaluate)
```

When #124 comes in, I suspect we'll want to expand on `test_evaluate_node` so that it encompasses checks for the new node types in #124.

src/causalprog/graph/node/base.py
```diff
@@ -127,6 +127,22 @@ def sample(
 
         """
 
+    @abstractmethod
+    def evaluate(
+        self,
+        **given_values: dict[str, float | npt.NDArray[float]],
+    ) -> float | npt.NDArray[float]:
+        """
+        Evaluate the node.
```
willGraham01:
```suggestion
        Evaluate the node, given evaluations of its precursor nodes.
```

willGraham01 (changes requested):
Some expansions to the test suite might be needed, and we probably shouldn't check for strict types when `evaluate`-ing since that opens a whole plethora of options we need to account for.

tests/test_algorithms/test_evaluate.py
```diff
@@ -0,0 +1,28 @@
...
+
+from causalprog.graph import DataNode
+
+
+@pytest.mark.parametrize(
+    ("node", "kwargs_to_evaluate", "expected_result"),
+    [
+        pytest.param(DataNode(label="A"), {"A": 2.0}, 2.0, id="Evaluate DataNode itself"),
+        # Likely want a test here for a ComponentNode?
+        pytest.param(ComponentNode("Parent", 1, label="Child"), {"Parent": np.arange(4)}, 1.0, id="Evaluate ComponentNode, given parent"),
```
mscroggs:
```suggestion
        pytest.param(DataNode(label="A"), {"A": 2.0}, 2.0, id="Evaluate DataNode itself"),
        pytest.param(ComponentNode("Parent", 1, label="Child"), {"Parent": np.arange(4)}, 1.0, id="Evaluate ComponentNode, given parent"),
```

src/causalprog/graph/node/component.py
mscroggs: Removed the check in [2c23e01](https://github.com/UCL/causalprog/pull/141/commits/2c23e019bdc433fd61f8ffb956928ab7d92d046e)

src/causalprog/graph/node/data.py
mscroggs: Resolved in [bb1b0f3](https://github.com/UCL/causalprog/pull/141/commits/bb1b0f3aea1412348fbb77c163a157598ce7d0f9)

willGraham01 (approved):

PR #142, opened 2026-06-24
Simplify some graph code

mscroggs:
- Added ConstantNode
- Replaced constant parameters with constant nodes
- Removed `is_parameter` and `is_distribution` from nodes
- Renamed `parameters` method to `parents`

Builds on top of #141


src/causalprog/graph/node/distribution.py
```diff
@@ -101,13 +101,8 @@ def __repr__(self) -> str:
...
     @property
-    def constant_parameters(self) -> dict[str, float]:
-        return self._constant_parameters
-
-    @override
-    @property
-    def parameters(self) -> dict[str, str]:
-        return self._parameters
+    def parents(self) -> list[str]:
+        return list(self._parameters.keys()) + list(self._constant_parameters.keys())
```
willGraham01:
```suggestion
        return [*self._parameters.keys(), *self._constant_parameters.keys()]
```

Saves creating two separate lists, then concatenating.

src/causalprog/graph/graph.py
```diff
@@ -91,22 +94,22 @@ def add_edge(self, start_node: Node | str, end_node: Node | str) -> None:
         self._graph.add_edge(start_node, end_node)
 
     @property
-    def parameter_nodes(self) -> tuple[Node, ...]:
+    def leaf_nodes(self) -> tuple[Node, ...]:
```
willGraham01: Leaf nodes have no child nodes, but here you're finding all the nodes that don't have parents. So aren't these the root nodes (have no parents), since they only have edges directed into other nodes? I could be wrong as by graph theory lingo is not super refreshed.

It is kind of weird that we have multiple root nodes, (but we expect to have multiple `DataNode` & `ConstantNode`s so not surprising).

src/causalprog/graph/graph.py
```diff
@@ -226,18 +218,19 @@ def model(self, **parameter_values: npt.ArrayLike) -> dict[str, npt.ArrayLike]:
 
         """
         # Confirm that all `DataNode`s have been assigned a value.
-        for node in self.parameter_nodes:
+        for node in self.leaf_nodes:
```
willGraham01: Just adding to my above comment on leaf vs root; on this line I definitely expected to read "`for node in self.root_nodes`" because "root implies that's where we start".

willGraham01 (changes requested):
Just need to clear up (either in discussion here if I'm wrong, or in the code if we want to switch) the "leaf vs root" question.

src/causalprog/graph/graph.py
mscroggs: ~We should discuss root vs leaf in our catch up this morning. I keep flipping the graph upside down and mixing up the terminology... Also our graphs aren't necessarily trees so leaves if probably the wrong terminology anyway~

mscroggs: ... actually having read the definitions of the terminology, you're right these should be roots. Fixing now

willGraham01 (approved):

Issue #139, opened 2026-06-04
Nice formats for the MLP functions


mscroggs:
The f functions that go into #124 can be messy. We need a nice way to build them.

(What follows below is text that @willGraham01 has added).

To expand on this, we essentially need a "nice builder" for the functions that we're going to attach to the `RVNodes` (see #124). To start with, we'll focus on Ricardo's specific requests for the MLPs we'll be encountering.

## Some MLP Notation Setting

An MLP $f_a$ is assumed to have $H_a$ hidden layers and $N_a$ units per hidden layer.

- I don't know if $N_a$ is a constant or a vector indicating different numbers of nodes per layer, but hopefully that's not too important a detail.

Given $H_a$ and $N_a$, $f_a$ is now entirely paramterised by the "model weights" or "model parameters" that define the maps between the layers / neurons in the MLP. We will call the collection of these parameters $\theta_a$.

Typically, one would then provide some training dataset of inputs $Z$ and outputs $Y$ to $f_a$, from which to "learn" the optimal model parameters $\theta_a = \theta_a^*$. After doing so, $f_a$ can be used to make predictions about the value of $Y$ for unseen evaluation points $Z = z$, effectively by calling

$$ f_a(z; \theta_a = \theta_a^*). $$

## What We Want

We need to have is the "function" $f_a(z; \theta_a)$ available to us as a Python callable; this issue requires us to provide a "nice way" to do this in our package. For example, we want to have a function/constructor/etc `mlp` in our package, usable in the following way:

```python
f_a = mlp(h_a, n_a)
```

The returned `f_a` needs to be callable object (function, class with `__call__`, etc) that can be used to evaluate $f_a(z; \theta_a)$  as described above:

```python
f_a(input_values, model_parameters)
```

which will ensure that it's compatible with our envisioned `.evaluate` algorithm (#125). 

The idea behind this is that we are going to attach MLPs to the `RVNodes` (#124) in our graph. For a node $X$ in a graph $G$, we will attach an MLP $f_X$ that is parameterised by a set of parameters $\theta_X$.

- (A vector or dictionary of ) The values of the precursor `RVNodes` of $X$ are the `input_values` to $f_X$.
- The values of the precursor `DataNode`s (#123) of $X$ are the `model_parameters` $\theta_X$.

Crucially, this means that we can use `evaluate` (#125) to propagate through a graph;

```python
theta = {
    "theta_X": ... # whatever values theta_X takes
}

[...]

PR #143, opened 2026-06-25
Add random variable nodes

### LINKED ISSUE #124: RVNodes

mscroggs:
Added nodes that represent random variables. These nodes then can be evaluated if the values of all of their parents are known/input.

This PR includes adding an assertion that labels are valid Python variable names, so that `**kwargs` interfaces can be used to set the values of nodes.

Resolves #124.


src/causalprog/graph/node/random_variables.py
```diff
@@ -0,0 +1,146 @@
...
+            shape: The shape of the output of the RV
+            label: A unique label to identify the node
+            compute: A function to compute node's value from given values of parents
+
+        """
+        super().__init__(label=label, shape=shape)
+        if compute is None:
+            self._parents = []
+        else:
+            self._parents = list(inspect.signature(compute).parameters.keys())
```
willGraham01, line 38: For now this is OK, but going to flag this because I imagine the functions created by #139 will not be able to be created with specific argument names, so the signature inspection won't work going forward (I could be wrong though).

In which case, we would require an input dictionary similar to what we had before, that translates node-labels to `compute` argument names.

Something that might just be more scalable is if we assume that `compute` has the same function signature as the functions $f_a$ in #139. We might have to do some manual packing-and-unpacking of various function arguments, but it would also avoid this problem.

src/causalprog/graph/node/random_variables.py
```diff
@@ -0,0 +1,146 @@
...
+        **given_values: float | npt.NDArray[float],
+    ) -> float | npt.NDArray[float]:
+        if self.label in given_values:
+            value = given_values[self.label]
+            if not self.is_valid_value(value):
+                msg = (
+                    f"Invalid value for {self.__class__.__name__}: "
+                    f"{self.label} cannot be {value}"
+                )
+                raise ValueError(msg)
```
willGraham01: I can imagine this check coming up quite a few times. Why not move it into the `is_valid_value` method, and allow sub-classes to call `__super__().is_valid_value()` after implementing their specific behaviour? You could give the method a `throw: bool` argument to toggle if you want the error to throw or the function to return `False`.

Alternatively, we could also make the case for not worrying about checking this here. If the user wants to input the wrong values into their own model, they should expect it to be broken?

src/causalprog/graph/node/random_variables.py
```diff
@@ -0,0 +1,146 @@
...
+            if self.shape != (value.shape if hasattr(value, "shape") else ()):
+                msg = f"Invalid value for node: {self.label}"
+                raise ValueError(msg)
```
willGraham01: Likewise, this could go into `is_valid_value` logic too, right?

src/causalprog/graph/node/random_variables.py
```diff
@@ -0,0 +1,146 @@
...
+            return value
+
+        if self._compute is None:
+            msg = f"Missing input for node: {self.label}."
+            raise ValueError(msg)
```
willGraham01, line 64: We should probably fail-fast and put this at the top of the method, right? Checking for `None` is going to be a lot faster than checking array shapes.

Also, the error message appears to be misleading. Maybe "Missing compute function for node {self.label}" is better?

src/causalprog/graph/node/random_variables.py
```diff
@@ -0,0 +1,146 @@
...
+        return self._compute(**{p: given_values[p] for p in self._parents})
```
willGraham01, line 65: 👍 Subject to call-changes based on how easy #139 is to implement.

tests/test_node/test_random_variables.py
```diff
@@ -0,0 +1,35 @@
...
+        node.evaluate()
+
+
+def test_invalid_discrete_node_value(raises_context):
+    node = DiscreteRandomVariableNode(label="Y", values=[-0.5, 0.0, 0.5])
+    with raises_context(ValueError("Invalid value for DiscreteRandomVariableNode")):
+        node.evaluate(Y=-1.0)
+
+
+def test_sum_parents():
```
willGraham01:
```suggestion
def test_evaluate_down_graph():
```

Slightly more expressive? Also we should probably cover a few more bases:

- Evaluate a node that is a child of 2 root nodes
- Evaluate a node that is a direct child of a root node, and a 2nd-generation child of another root node (to check that evaluation through the graph works OK).

You can probably come up with a single `graph` instance (to use as a fixture) that can encompass these two cases and the one you already have.

willGraham01 (commented):
Largely OK but a few comments on some naming conventions, and flagging that the signature of `compute` might not be able to be taken for granted. But for now we can ignore this until we have something concrete for #139, and then look at getting the two of them to play nicely.

src/causalprog/graph/node/random_variables.py
mscroggs: Agreed, but I propose that we leave this for now and edit later if/when we hit an issue.

src/causalprog/graph/node/random_variables.py
mscroggs: I've added `assert_is_valid_value` that uses `is_valid_value` as I'm not sure I like having bools to control whether the function returns or panics

src/causalprog/graph/node/random_variables.py
mscroggs: I've moved it into `assert_is_valid_value`

src/causalprog/graph/node/random_variables.py
mscroggs: The idea is that if the value is given for this distribution, then compute is ok being None (as we don't need to compute anything). The checking only happens if this node is in the given values, so I think this is ok

tests/test_node/test_random_variables.py
mscroggs: The further tests will come in #145 where I've added the extra code to traverse the graph and evaluate as we go

src/causalprog/graph/node/random_variables.py
mscroggs: ... but in general, I prefer making the user be explicit rather than using magic, so will like do a PR to replace this signature checking with a parents input, but I think this is ok for now to get us started

willGraham01: Agreed with this - something to come back to for now. We might find that `jax`'s hunger for `PyTrees` means we need to use `dict`s anyway, in which case this will be a moot point.

src/causalprog/graph/node/random_variables.py
willGraham01: Yeah my bad - I read this before reading [your Slack comment](https://ucl-arc.slack.com/archives/C089BQ4DF8R/p1782481100020039?thread_ts=1782478625.657899&cid=C089BQ4DF8R) and coming round to what you already had. Ignore me :sweat_smile:

willGraham01 (approved):

PR #145, opened 2026-06-26
Add evaluate algorithm

### LINKED ISSUE #125: Add evaluate algorithm
- Inputs: graph, node label, values for data nodes
- Returns: value at node with label

This should:
- Iterate using `roots_down_to_outcome`:
    - Evaluate and store result

Note: this is very similar to existing sample in algorithms.moments

### LINKED ISSUE #126: Helper function to build graph Ricardo is interested in
This but with more thetas:

Image

This graph has:
- DataNodes: L, Z, $\theta_X$, $\theta_r$, $\theta_\pi$, $\theta_m$, $\theta_y$, $\theta_r$
- DiscreteRVNodes: C, takes values from 1 to K (uniformly for now)
- ContinuousRVNodes:
    - UX (parents C, L)
    - X (parents UX, Z, L, $\theta_X$)
    - UY (parents C, UX, $\theta_\pi$, $\theta_r$, $\theta_m$, L?)
    - Y (parents X, UY, $\theta_y$, L?)

As inputs, user has to specify structure for f functions.

mscroggs:
- Add `evaluate` algorithm that uses the `evaluate` method on each node from the roots down to the outcome node.
- Add a function to `utils` that creates the graph for the example model that Ricardo is interested in

Builds on #143.

Resolves #125. Resolves #126.


src/causalprog/graph/special.py
```diff
@@ -0,0 +1,80 @@
...
+
+from causalprog.graph import (
+    ContinuousRandomVariableNode,
+    DataNode,
+    DiscreteRandomVariableNode,
+    Graph,
+)
+
+
+def example_model(
```
mscroggs, line 14: Do we think `causalprog.utils.graphs` is a good place for this function or should we put it elsewhere?

tests/test_utils/test_graphs.py
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    assert np.allclose(evaluate(graph, "L", **data), np.array([5.5]))
+    assert np.allclose(evaluate(graph, "Z", **data), np.array([2.0]))
+    assert np.allclose(evaluate(graph, "C", **data), 4.0)
+    assert np.allclose(evaluate(graph, "UX", **data), 4.0)
+    assert np.allclose(evaluate(graph, "UY", **data), 5.0)
+    assert np.allclose(evaluate(graph, "PhiX", **data), 5.5)
+    assert np.allclose(evaluate(graph, "X", **data), 0.5)
+    assert np.isclose(evaluate(graph, "Y", **data), 2.5)
```
mscroggs: I think this test hits all the cases in https://github.com/UCL/causalprog/pull/143#discussion_r3481111018.

Graph for reference:

Image

src/causalprog/algorithms/evaluate.py
```diff
@@ -0,0 +1,26 @@
...
+        outcome_node_label: The label of the node to evaluate.
+        values: Values taken by nodes whose value is given
+
+    Returns:
+        The evaluation of the node
+
+    """
+    for node in graph.roots_down_to_outcome(outcome_node_label):
+        if node.label not in values:
+            values[node.label] = node.evaluate(**values)
```
willGraham01:
```suggestion
    nodes_to_evaluate = [
        n for n in graph.roots_down_to_outcome(outcome_node_label)
        if n not in values
    ]
    for node in nodes_to_evaluate:
            values[node.label] = node.evaluate(**values)
```

List comprehensions preserve ordering, so this shouldn't change the evaluation order. But it means we can pre-compute the list to loop over rather than looping over everything and then checkng & halting when we encounter something we already have a value for.

src/causalprog/graph/special.py
willGraham01: Only other place I can see for it would be somewhere like `causalprog.graphs.static` or `causalprog.graphs.special`; IE somewhere under the `graphs` sub-module.

Since this is ultimately a graph builder, I think I have a slight preference for under the `graphs` sub-module but I won't die on this hill.

src/causalprog/graph/special.py
```diff
@@ -0,0 +1,80 @@
...
+    *,
+    label: str = "G",
+    l_len: int = 1,
+    z_len: int = 1,
+    k: int = 10,
+    compute_u_x: Callable,
+    compute_u_y: Callable,
+    compute_phi_x: Callable,
+    compute_x: Callable,
+    compute_y: Callable,
```
willGraham01, line 24: (Another internal reminder to revisit this once #139 is done, or we know what object is produced by that).

src/causalprog/graph/special.py
```diff
@@ -0,0 +1,80 @@
...
+            msg = f"Invalid input to PhiX: {p}"
+            raise ValueError(msg)
+    for p in inspect.signature(compute_x).parameters:
+        if p not in ["Z", "PhiX", "UX"]:
+            msg = f"Invalid input to X: {p}"
+            raise ValueError(msg)
+    for p in inspect.signature(compute_y).parameters:
+        if p not in ["X", "UY"]:
+            msg = f"Invalid input to X: {p}"
+            raise ValueError(msg)
```
willGraham01, line 63: Another subject to #139 comment (we might just ditch explicit naming of arguments and go for `dict` unpacking everywhere).

src/causalprog/algorithms/evaluate.py
```diff
@@ -0,0 +1,26 @@
+"""Algorithms for evaluating a graph node."""
+
+import numpy.typing as npt
+
+from causalprog.graph import Graph
+
+
+def evaluate(
+    graph: Graph, outcome_node_label: str, **values: float | npt.NDArray[float]
```
willGraham01, line 35: This is not [a pure function](https://docs.jax.dev/en/latest/notebooks/Common_Gotchas_in_JAX.html#pure-functions) (`values` is updated in-place) which will upset JAX.

But this is fairly easily fixed by simply doing `computed_values = {}` before the loop, and then doing `computed_values[node.label] = node.evaluate(**values, **computed_values)`.

You can then return `computed_values[outcome_node_label]`.

src/causalprog/algorithms/evaluate.py
```diff
@@ -0,0 +1,26 @@
...
+        values: Values taken by nodes whose value is given
+
+    Returns:
+        The evaluation of the node
+
+    """
+    for node in graph.roots_down_to_outcome(outcome_node_label):
+        if node.label not in values:
+            values[node.label] = node.evaluate(**values)
+    return values[outcome_node_label]
```
willGraham01: I can imagine that there will be cases when we'll want to return everything in `values`, not just the final node value (cf #129 for example). As it currently stands, we'd need one `evaluate` call per node we want to evaluate, even if we're evaluating them "on the way down".

Is it worth just returning `values` itself (or `computed_values` following on from my comment about pure functions)? Or alternatively, changing `evaluate` to be called as

```python
def evaluate(graph: Graph, *outcome_node_label: str, **values: npt.NDarray[float]) -> dict[str, npt.NDArray[float]]:
```

where some internal logic is done to ensure all `outcome_node_label`s are evaluated (should be as easy as doing `graph.roots_down_to_outcome` multiple times and appending the results whilst dropping duplicate entries), and then a dictionary mapping these `outcome_node_label`s to the corresponding evaluations is returned?

src/causalprog/algorithms/__init__.py
```diff
@@ -1,4 +1,5 @@
 """Algorithms."""
 
 from .do import do
+from .evaluate import evaluate
 from .moments import expectation, moment, standard_deviation
```
willGraham01, line 5:
```suggestion
from .moments import expectation, moment, standard_deviation

__all__ = ("do", "evaluate", "expectation", "moment", "standard_deviation")

```

tests/test_utils/test_graphs.py
willGraham01: I think so too, but I would argue that this is a test of the `algorithms.evaluate` method so it should live inside `test_algorithms/test_evaluate.py`, maybe as some of the parametrised test cases.

I reckon in this test you should just do a quick reality check that all the edges we expect are present (and pointing the right ways!)

tests/test_algorithms/test_evaluate.py
```diff
@@ -46,3 +53,27 @@ def test_evaluate_node_fail_on_missing_data(
...
+    graph = Graph(label="g")
+    graph.add_node(DataNode(label="a"))
+    graph.add_node(DataNode(label="b"))
+    graph.add_node(ContinuousRandomVariableNode(label="c", compute=lambda a: a - 0.5))
+    graph.add_node(
+        ContinuousRandomVariableNode(label="x", compute=lambda b, c: c + 2.0 * b)
+    )
+
+    assert np.isclose(evaluate(graph, "c", a=2.0, b=1.5), 1.5)
+    assert np.isclose(evaluate(graph, "x", a=2.0, b=1.5), 4.5)
```
willGraham01: As I mentioned elsewhere, these test cases can be parametrised and are in fact special cases of [what you're currently checking for Ricardo's graph](https://github.com/UCL/causalprog/pull/145/changes#diff-6c71aa83f4cc4d080d03b4b9a711b73bb9714533e1ba7ff0ad6d3edce297f41eR23-R30).

```python
from jax import Array
import jax.numpy as jnp
from causalprog.utils.graphs import example_model


@pytest.fixture
def evaluate_test_graph() -> Graph:
    return example_model(
        z_len=2,
        compute_u_x=lambda C: C,
        compute_u_y=lambda C: C + 1,
        compute_phi_x=lambda L: L[0],
        compute_x=lambda Z, PhiX, UX: Z[0] + UX - PhiX,
        compute_y=lambda X, UY: X * UY,
    )


@pytest.mark.parametrize(
    ("outcome_node_label", "initial_values", "expected_result"),
    [
        pytest.param(
            "L",
            {"L": jnp.array([5.5]), "Z": jnp.array([2.0, 0.0]), "C": 4.0},
            jnp.array([5.5]),
            id="DataNode evaluation w/ excess information provided",
        ),
        pytest.param(
            "Z",
            {"Z": jnp.array([2.0, 0.0])},
            jnp.array([2.0, 0.0]),
            id="DataNode evaluation",
        ),
        pytest.param(
            "C",
            {"C": 4.0},
            4.0,
            id="DiscreteRVNode evaluation",
        ),
        pytest.param(
            "UX",
            {"C": 4.0},
            4.0,
            id="CtsRVNode evaluation",
        ),
        pytest.param(
            "UX",
            {"C": 4.0, "UX": 1.0},
            1.0,
            id="CtsRVNode evaluation, 'given that' overrides computed value",
        ),
        pytest.param(
            "UY",
            {"C": 4.0},
            5.0,
            id="CtsRVNode evaluation, with parents that need evaluating",
        ),
        pytest.param(
            "X",
            {"L": jnp.array([5.5]), "Z": jnp.array([2.0, 0.0]), "C": 4.0},
            0.5,
            id="Multiple paths from different root nodes",
        ),
        pytest.param(
            "X",
            {"L": jnp.array([5.5]), "Z": jnp.array([2.0, 0.0]), "C": 4.0, "PhiX": 0.0},
            6.0,
            id="Multiple paths from different root nodes, with some given values",
        ),
        pytest.param(
            "Y",
            {"L": jnp.array([5.5]), "Z": jnp.array([2.0, 0.0]), "C": 4.0},
            2.5,
            id="Evaluating the 'outcome' node.",
        ),
    ],
)
def test_evaluate(
    evaluate_test_graph: Graph,
    outcome_node_label: str,
    initial_values: dict[str, Array],
    expected_result: Array,
) -> None:
    computed_result = evaluate(
        evaluate_test_graph, outcome_node_label, **initial_values
    )
    # If we start returning dicts, will need a loop over the elements here.
    assert jnp.allclose(expected_result, computed_result)
```

This parametrisation also makes the current `test_evaluate_algorithm_{three,four}_node` and `test_evaluate_node` tests redundant as it absorbs the cases that they're examining.

tests/test_algorithms/test_evaluate.py
willGraham01: This can be generalised to catch all more error cases; specifically when a `DiscreteRVNode` is evaluated using an invalid value.

willGraham01 (changes requested):
Main comment is about how `evaluate` is currently implemented; namely

- It's not a pure function and
- We might want to retrieve multiple values from the graph, some of which are "on the way down". Right now we have to do this through multiple calls.

Beyond that, some comments on organising and expanding the tests.

src/causalprog/graph/special.py
willGraham01: Something else I noticed just now is that here you have underscores in the `compute...` names, but the labels use `CamelCasing` without underscores.

src/causalprog/graph/special.py
mscroggs: I'm moving it to `causalprog.graphs.special`. I think that makes more sense

src/causalprog/graph/special.py
mscroggs: Yeah, that's caused by indecision about how much to appease ruff

src/causalprog/algorithms/evaluate.py
mscroggs: good spot, I'd missed that values was mutable

src/causalprog/algorithms/evaluate.py
mscroggs: I've added `evaluate_down_to` which does the same as this but return the dict of all values (and updated `evaluate` to call that and return the value at the node of interest. I guess we can revisit exactly how this interface looks once we get to writing some example notebooks

src/causalprog/graph/special.py
mscroggs: I added a test of the edges in https://github.com/UCL/causalprog/pull/145/changes/9d65fcc3a03c783d5ba4db0e3e8ee36d95375304

tests/test_algorithms/test_evaluate.py
mscroggs: I'm going to add tests of invalid values in a separate branch

src/causalprog/algorithms/evaluate.py
```diff
@@ -0,0 +1,49 @@
...
+    Args:
+        graph: The graph that the node is contained in.
+        outcome_node_label: The label of the node to evaluate down to.
+        values: Values taken by nodes whose value is given
+
+    Returns:
+        A dictionary of the values of all the nodes that are ancestors of the input node
+
+    """
+    computed_values = dict(values)
```
willGraham01: `values` could contain some pretty hefty arrays (and if so, the user already has access to those arrays in `values`). So maybe just initialise as an empty dictionary (and then we also get the side effect of only returning new, previously unknown node values too).

If you do this, you'll need to double-unpack in `node.evaluate` of course:

```python
computed_values[node.label] = node.evaluate(**values, **computed_values)
```

src/causalprog/graph/special.py
willGraham01: No problem, for now let's leave as-is and postpone the style decisions to later 😅

tests/test_algorithms/test_evaluate.py
```diff
@@ -1,48 +1,92 @@
...
-def test_evaluate_node_fail_on_missing_data(
-    node, kwargs_for_evaluate, expected_error, raises_context
-):
-    with raises_context(expected_error):
-        node.evaluate(**kwargs_for_evaluate)
+def test_evaluate(
+    evaluate_test_graph: Graph,
+    outcome_node_label: str,
+    initial_values: dict[str, Array],
+    expected_result: Array,
```
willGraham01: I hate to be a pain but we should probably also have a sanity check for `evaluate_down_to` here too. But I think we can do this inside this test:

- Change the `expected_result` parameter value to be the dictionary that we expect to come out of `evaluate_down_to`.
- `computed_result` should be the output of `evaluate_down_to`.
- `computed_result_singular = evaluate(...)` to store the result of explicitly calling `evaluate`.
- Then:

```python
# Same number of entries
assert len(expected_result) == len(computed_result)
# Keys are correct
assert set(expected_result.keys()) == set(computed_result.keys())

# All entries match to acceptable precision for floats
for node_label, computed_value in computed_result.items():
    assert jnp.allclose(computed_value, expected_result[node_label]

# Just asking for one value did indeed extract the correct node value
assert jnp.allclose(computed_result_singular, computed_result[outcome_node_label])
```

willGraham01 (changes requested):
`evaluate_down_to` is more general than `evaluate` so we should adapt the test accordingly (though we can test both in the same test because I'm not a code sadist).

Also we might not want to duplicate the `values` dictionary entries, so have suggested just returning node-evaluations that are actually computed.

src/causalprog/algorithms/evaluate.py
mscroggs: Fixed in [1b92a35](https://github.com/UCL/causalprog/pull/145/commits/1b92a35093dca66823a6701cf16ad7974d4d2bbe)

tests/test_algorithms/test_evaluate.py
mscroggs: [c0c54f8](https://github.com/UCL/causalprog/pull/145/commits/c0c54f8698ec2ec4168b50c2be89c5135a24c78e)

tests/test_algorithms/test_evaluate.py
```diff
@@ -1,48 +1,109 @@
...
```
willGraham01:
```suggestion
    expected_result: dict[str, float | Array],
```

(Soz, but otherwise all looks good!)

willGraham01 (approved):

tests/test_algorithms/test_evaluate.py
mscroggs: We should work out why mypy didn't spot that one

willGraham01: Haven't we just blanket disabled typehint checking in the tests (because a lot of our fixtures have very long types)?

PR #146, opened 2026-06-26
Add `QuadratureMethod` for Ricardo's use case

willGraham01:
Concerns #129 |

Implements the `UniformWeightMonteCarloGaussianQuadrature` class, which performs Monte-Carlo integration using Gaussian samples, but uniform weights. We intend to use this in the regression function (builder) needed by #129.

Mathematically, this scheme is identical to performing standard Monte-Carlo integration (with Gaussian importance sampling), but absorbing the "sample weight function" / "PDF of the importance sampling distribution" inside the integrand itself. That is to say, if we were to use `UniformWeightMonteCarloGaussianQuadrature` to integrate a function $f$ over $[a, b]$, we would be doing

$$
\int_a^b f(x) dx \approx \frac{1}{N}\sum_i^N f(x_i),
$$

where $x_i$ are $N$ samples drawn from a truncated normal distribution on $[a,b]$. If $\mathcal{P}(x; a, b)$ represents the PDF of a truncated normal distribution on $[a, b]$, and we were instead to conduct standard Monte-Carlo integration with Gaussian importance sampling (as per the `MonteCarloGaussianQuadrature` class), note that we would have

$$
\int_a^b f(x)\mathcal{P}(x; a, b) dx \approx \frac{1}{N}\sum_i^N f(x_i).
$$

As such; integrating $f$ using `UniformWeightMonteCarloGaussianQuadrature` is the same as integrating $f\mathcal{P}$ using `MonteCarloGaussianQuadrature`. Or alternatively, integrating $f$ using `UniformWeightMonteCarloGaussianQuadrature` is the same as computing $\mathbb{E}[f(X)]$ where $X\sim\mathcal{P}(\cdot; a, b)$ (in an abuse of notation).


src/causalprog/quadrature/__init__.py
```diff
@@ -1,4 +1,13 @@
...
+from .monte_carlo import (
+    MonteCarloGaussianQuadrature,
+    UniformWeightMonteCarloGaussianQuadrature,
+)
+
+__all__ = (
+    "GaussianQuadrature",
+    "MonteCarloGaussianQuadrature",
+    "UniformWeightMonteCarloGaussianQuadrature",
+)
```
mscroggs, line 13: Why is this `__all__` needed?

mscroggs (approved):

src/causalprog/quadrature/__init__.py
willGraham01: 1. Appease ruff
2. If a bad user does `from causalprog.quadrature import *`, the things that are imported are only the things whose names appear in `__all__`, rather than polluting the entire namespace with the various sub-modules etc in this `__init__`'s scope.

PR #147, opened 2026-06-29
Regression Function Builder

willGraham01:
Progresses #129 |

Adds the `build_regression_function` function, which assembles the regression function $r$ in #129 for Ricardo's graph use case.

Note that a lot of "morphing" was done in this PR, so things look very messy. In particular:

- `causalprog/graph/special.py` was renamed to `causalprog/graph/ricardo.py.
  - This is where `build_regression_function` was added, to keep things in one place.
- In order to appease `mypy`, the `._compute` method of `RandomVariableNode` needed to be exposed.
  - This is now done via a simple wrapper property, `.compute`.
  - However, this necessitates us being unable to pass in `compute = None` when creating `RVNode`s, which is an issue since in the vast majority of tests we _don't_ actually need this :upside_down_face: Solution was to add a default function, `_compute_not_set`, that `.compute` is set to if nothing else is given. This means that `mypy` is happy, we can create `RVNode`s without needing to pass a bunch of un-necessary `lambda`s around when we don't need to, _and_ we get a sensible error thrown if trying to call `.compute` when it's not been set.
- `r_builder` can only be used with the `UniformWeightMonteCarloGaussianQuadrature` for evaluating it's integral, but #129 requires that we are also able to use `GaussianQuadrature` too.
  - As it currently stands, the method will break if we try to use `GaussianQuadrature` since we are assuming the domain of integration is the whole real line.
  - For now, `build_regression_function` just throws an error if the given quadrature scheme isn't a `UniformWeightMonteCarloGaussianQuadrature`. Will address this in another PR - presumably we would have to ask for the domain of integration when calling `build_regression_function` and pass this to the method. Then it's the user's fault if they try to use `GaussianQuadrature` on an infinite domain.
- Enabling `x64` precision is necessary for one of the two test cases to pass. Bizarrely, enabling this setting then causes one of the _subsequent_ tests to fail :upside_down_face:
  - See relevant [logs here](https://github.com/UCL/causalprog/actions/runs/28657062246/job/84988616865#step:6:135) - the test that is failing is unrelated to this PR.
  - Got around this by adding a `pytest.fixture` that enables activation of this setting for a single test at a time (by switching it on, running the test, then switching it off).
  - It would be nicer if this hack could be a decorator, rather than a `pytest.fixture` (since otherwise ruff complains we're not "using" the argument) but I couldn't figure out how to write decorators that also worked with test functions (using a normal decorator seems to prevent fixture discovery - or at least passing of fixture values into the decorated function).

Besides the `build_regression_function`, I managed to come up with two test cases, whose docstrings hopefully explain what's going on. Essentially they both reduce the problem to something somewhat trivial but in a manner which means we can analytically verify that the regression function is being built (or at least evaluated) in the manner we expect.


src/causalprog/graph/node/random_variables.py
```diff
@@ -13,6 +13,25 @@
 class RandomVariableNode(Node):
     """A node containing a random variable (RV)."""
 
+    def _compute_not_set(self, *args, **kwargs) -> None:
```
mscroggs, line 16: Could this be a function in the main file rather than in the class so that it doesn't show up as a method to anyone doing `node._` then hitting tab in ipython?

src/causalprog/graph/node/random_variables.py
```diff
@@ -63,6 +83,11 @@ def evaluate(
             raise ValueError(msg)
         return self._compute(given_values)
 
+    @property
+    def compute(self) -> typing.Callable:
+        """Direct call to the `._compute` attribute."""
+        return self._compute
+
```
mscroggs: Would this be neater than returning a callable?

```suggestion
    def compute(self, *args, **kwargs) -> {{TYPE}}:
        """Direct call to the `._compute` attribute."""
        return self._compute(*args, **kwargs)

```

mscroggs: where `{{TYPE}}` is maybe something like `float | np.NDArray[float]`?

tests/test_graph/test_ricardo.py
```diff
@@ -1,4 +1,4 @@
-from causalprog.graph.special import example_model
+from causalprog.graph.ricardo import example_model
```
mscroggs, line 1: This test file should probably be renamed to test_ricardo.py to match the move

src/causalprog/graph/ricardo.py
```diff
@@ -0,0 +1,189 @@
...
+    - The node $U_Y$ stores the function $\pi_{ul}(c)$ in it's `.compute` attribute.
+      $U_Y$ also provides access to the functions $f_r$ and $f_m$ through two of its
+      attributes, and has two nodes representing $\theta_r$ and $\theta_m$ as parents.
+    """
+    if not isinstance(quadrature, UWMCGQuad):
+        msg = (
+            "Only UniformWeightMonteCarloGaussianQuadrature "
+            "is supported as a quadrature method."
+        )
+        raise TypeError(msg)
```
mscroggs: Maybe a `NotImplementedError` to suggest this will be supported in future?

src/causalprog/graph/ricardo.py
```diff
@@ -0,0 +1,189 @@
...
+        u = f_x_inverse(xzl)
+
+        theta_m = model_params["theta_m"]
+        theta_r = model_params["theta_r"]
+        theta_pi = model_params["theta_pi"]
+        theta_y = model_params["theta_y"]
+
+        x = xzl["x"]
+        z = xzl["z"]
+        el = xzl["l"]
```
mscroggs, line 156: To consider probably in a separate PR: `"l"` is maybe a bad label (for same reason `l = ` here would upset ruff: `"l"` and `"I"` are indistinguishable in many fonts). I don't know what alternative label would be nice here though

mscroggs: I opened #153 for this so we can ignore this comment for now

src/causalprog/graph/ricardo.py
```diff
@@ -0,0 +1,189 @@
...
+
+        result = 0.0
+        for i_c, c in enumerate(c_values):
+            czl = {"c": c, "z": z, "l": el}
```
mscroggs, line 160: Is this dictionary essentially `xzl` but with some entries removed? Could we input `xzl` instead to simplify this?

tests/test_integration/test_ricardo_regression_function.py
```diff
@@ -0,0 +1,215 @@
...
+        k=k_len,
+        z_len=z_len,
+        compute_u_x=f_ux,
+        compute_u_y=f_pi,
+        compute_phi_x=None,
+        compute_x=None,
+        compute_y=f_y,
+    )
+    # Manually attach methods to node for now. FIXME: should be removed once we have
+    # a more elegant solution for attaching additional functions to nodes.
```
mscroggs, line 27: Do we want an issue for this FIXME?

mscroggs (approved):
Approving as happy for this to be merged once my minor comments are looked at

src/causalprog/graph/node/random_variables.py
willGraham01: It needs to implicitly receive `self` as the first argument, before `*args` and `**kwargs`. I couldn't find an elegant solution that didn't involve also warping how `evaluate` and various other methods behaved.

Taking outside the class means that it no longer receives the `Node` instance as the implicit first argument, which in turn means that there's no way to inform the user of which node is throwing the error (`self.label` isn't available). We could just use the error message "compute not set" but without telling the user which node is the problem, this seems somewhat unhelpful.

We also can't just add the node label as a positional argument either, since the signature of the method needs to be consistent with how it's expected to be called in `evaluate` (where it will only be passed dictionaries of previous node values).

I'd also argue that anyone doing `node._` `[TAB]` isn't being a good person since they're deliberately looking for "private" methods 😂

src/causalprog/graph/node/random_variables.py
willGraham01: Indeed it would. Also swapped the internals of `evaluate` to use `self.compute` over `self._compute` just to be a bit cleaner.

src/causalprog/graph/ricardo.py
willGraham01: It has `c` as an extra key, and drops `x`. I guess I could do `{"c": c, **xzl}` if we want to? 

Will leave it up to you as to whether that's nicer than what I've got here already.

tests/test_integration/test_ricardo_regression_function.py
willGraham01: Yeah we probably do since this is something Ricardo wants to do as part of his particular graph.

Now covered in #156. It's actually something we should probably address in phase 3, but it is *technically* optional.

src/causalprog/graph/ricardo.py
mscroggs: Ah, I misread `"c": c` as `"x": x` so they looked the same. I think this is fine

PR #148, opened 2026-06-29
Add check that values input into evaluate are valid

mscroggs:
Adds test that suggested in comments #145, and update code to make the test pass.


src/causalprog/graph/node/base.py
```diff
@@ -168,3 +168,19 @@ def parents(self) -> list[str]:
             List of labels of parent nodes
 
         """
+
+    @abstractmethod
+    def is_valid_value(self, value: float | npt.NDArray[float]) -> bool:
```
willGraham01: This function just returns `True` for every class except discrete random variable nodes... maybe we can afford to provide a default implementation for our own sanity?

`@override` will still work when we do want to change the behaviour.

src/causalprog/algorithms/evaluate.py
```diff
@@ -46,6 +48,8 @@ def evaluate(
         The evaluation of the node
 
     """
+    for label, value in values.items():
+        graph.get_node(label).assert_is_valid_value(value)
     if outcome_node_label in values:
         return values[outcome_node_label]
     return evaluate_down_to(graph, outcome_node_label, values)[outcome_node_label]
```
willGraham01: This can be moved inside `evaluate_down_to` to avoid duplication, right?

```python
def evaluate_down_to(...):

    for label, value in values.items():
        graph.get_node(label).assert_is_valid_value(value)
    if outcome_node_label in values:
        return {outcome_node_label: values[outcome_node_label]}

    # Rest of method remains as is.

def evaluate(...):
    return evaluate_down_to(...)

willGraham01 (changes requested):
Think we can save on code additions in a few places.

src/causalprog/algorithms/evaluate.py
mscroggs: The `if outcome_node_label` made it hard, but I think your suggestion sorts it as long at that won't be copying large data (I wish Python was a bit more explicit about sometimes secretly copying data)

src/causalprog/graph/node/base.py
mscroggs: Yes, that would be simpler

src/causalprog/algorithms/evaluate.py
willGraham01: > I wish Python was a bit more explicit about sometimes secretly copying data

We should re-write the whole thing in C++, naturally :grin: 

And I think this assignment (`{outcome_node_label: values[outcome_node_label]}`) in fact won't make a copy of `values[outcome_node_label]`, but rather use a reference? Though I could be wrong, but regardless this is a post-getting-things-to-run optimisation problem.

willGraham01 (approved):

PR #149, opened 2026-06-29
Replace kwargs input in evaluate with a dict

mscroggs:
- Replace `**values` kwargs input with a dict input.
- Add explicit `parents` input to random variable nodes as we can no longer use function signature magic.
- Remove assertion that labels must be valid Python variable names as they no longer need to be valid kwargs keys.
- Remove ignoring of ruff warning about non-lowercase function arguments in tests

We should review #150 before merging this.


tests/test_algorithms/test_evaluate.py
```diff
@@ -13,11 +13,11 @@
...
-        compute_u_x=lambda C: C,
-        compute_u_y=lambda C: C + 1,
-        compute_phi_x=lambda L: L[0],
-        compute_x=lambda Z, PhiX, UX: Z[0] + UX - PhiX,
-        compute_y=lambda X, UY: X * UY,
+        compute_u_x=lambda data: data["C"],
+        compute_u_y=lambda data: data["C"] + 1,
+        compute_phi_x=lambda data: data["L"][0],
+        compute_x=lambda data: data["Z"][0] + data["UX"] - data["PhiX"],
+        compute_y=lambda data: data["X"] * data["UY"],
```
mscroggs: We still have the issue here that the node labels are uppercase, but the function names are lowercase...

mscroggs: #150 changes the labels to also be lowercase. We should merge that into this branch before merging into main if we agree with moving all to consistent lowercase

willGraham01 (approved):

PR #151, opened 2026-07-02
Add replace_node algorithm

mscroggs:


tests/test_algorithms/test_replace_node.py
```diff
@@ -0,0 +1,60 @@
...
+    assert edges == {
+        ("l", "phi_x"),
+        ("c", "u_y"),
+        ("c", "u_x_updated"),
+        ("x_updated", "u_x_updated"),
+        ("phi_x", "x_updated"),
+        ("z", "x_updated"),
+        ("u_y", "y"),
+        ("x_updated", "y"),
+    }
```
mscroggs: This test tests the reversing of the edge u_x -> x in Ricardo's graph. It would probably be good to add smaller unit tests too...

src/causalprog/algorithms/replace_node.py
```diff
@@ -0,0 +1,36 @@
...
+        node: The new node to be inserted.
+        label: The label of the new graph.
+
+    Returns:
+        A copy of the graph with the replacement made.
+
+    """
+    if label is None:
+        label = f"{graph.label}_updated"
+    g = graph.copy()
```
willGraham01: This logic appears twice, and you aren't passing the user-provided label (if it exists) into `graph.copy`.

Did you mean to just do


```suggestion
    g = graph.copy(label=label)
```

since `graph.copy()` handles the case `label=None` explicitly inside it anyway?

src/causalprog/algorithms/replace_node.py
```diff
@@ -0,0 +1,36 @@
+"""Algorithm for replacing a graph node."""
+
+from causalprog.graph import Graph, Node
+
+
+def replace_node(
+    graph: Graph, node_label: str, node: Node, label: str | None = None
```
willGraham01: These variable names make the logic of the method quite hard to follow.

- `node_label` -> `node_(label_)to_replace`
- `node` -> `replacement_node`
- `label` -> `new_graph_label` (since there are 2 labels floating around already!)

Might clear things up?

src/causalprog/graph/node/base.py
```diff
@@ -184,3 +184,7 @@ def assert_is_valid_value(self, value: float | npt.NDArray[float]) -> None:
         if self.shape != (value.shape if hasattr(value, "shape") else ()):
             msg = f"Invalid value for node: {self.label}"
             raise ValueError(msg)
+
+    @abstractmethod
+    def replace_parent(self, old_parent_label: str, new_parent_label: str) -> None:
+        """Replace a parent node."""
```
willGraham01: The same block of code appears whenever this method is overridden, since every `Node` subclass has `self.parents` defined (as a list) and you happen to be using this property's value (but not the property itself) in each of the subclass-specific implementations.

So I think the base class should just implement this method directly, and let it be extended with `super()` when necessary. 

```suggestion
    def replace_parent(self, old_parent_label: str, new_parent_label: str) -> None:
        """Replace a parent node."""
        if old_parent_label not in self.parents:
            msg = (
                "Cannot replace parent node: "
                f"{old_parent_label} is not a parent of {self.label}"
            )
            raise ValueError(msg)
```

My suggestions on the other classes are the changes that would be needed if we go ahead with this plan.

src/causalprog/graph/node/component.py
```diff
@@ -82,3 +82,13 @@ def __repr__(self) -> str:
...
+
+    @override
+    def replace_parent(self, old_parent_label: str, new_parent_label: str) -> None:
+        if old_parent_label != self._parent_node_label:
+            msg = (
+                "Cannot replace parent node: "
+                f"{old_parent_label} is not a parent of {self.label}"
+            )
+            raise ValueError(msg)
+        self._parent_node_label = new_parent_label
```
willGraham01, line 89:
```suggestion
    def replace_parent(self, old_parent_label: str, new_parent_label: str) -> None:
        super().replace_parent(old_parent_label, new_parent_label)
        self._parent_node_label = new_parent_label
```

src/causalprog/graph/node/data.py
```diff
@@ -67,3 +67,11 @@ def __repr__(self) -> str:
...
     def parents(self) -> list[str]:
         return []
+
+    @override
+    def replace_parent(self, old_parent_label: str, new_parent_label: str) -> None:
+        msg = (
+            "Cannot replace parent node: "
+            f"{old_parent_label} is not a parent of {self.label}"
+        )
+        raise ValueError(msg)
```
willGraham01:
```suggestion
```

(Since `self.parents == []` on this class, so no specific override is required).

src/causalprog/graph/node/constant.py
```diff
@@ -61,3 +61,11 @@ def __repr__(self) -> str:
...
     def parents(self) -> list[str]:
         return []
+
+    @override
+    def replace_parent(self, old_parent_label: str, new_parent_label: str) -> None:
+        msg = (
+            "Cannot replace parent node: "
+            f"{old_parent_label} is not a parent of {self.label}"
+        )
+        raise ValueError(msg)
```
willGraham01:
```suggestion
```

(Since `self.parents == []` on this class, so no specific override is required).

src/causalprog/graph/node/distribution.py
```diff
@@ -126,3 +126,15 @@ def create_model_site(self, **dependent_nodes: jax.Array) -> npt.ArrayLike:
...
         )
+
+    @override
+    def replace_parent(self, old_parent_label: str, new_parent_label: str) -> None:
+        if old_parent_label not in self._parameters.values():
+            msg = (
+                "Cannot replace parent node: "
+                f"{old_parent_label} is not a parent of {self.label}"
+            )
+            raise ValueError(msg)
```
willGraham01:
```suggestion
        super().replace_parent(old_parent_label, new_parent_label)
```

src/causalprog/graph/node/random_variables.py
```diff
@@ -68,6 +68,17 @@ def evaluate(
...
         return self._parents
 
+    @override
+    def replace_parent(self, old_parent_label: str, new_parent_label: str) -> None:
+        if old_parent_label not in self._parents:
+            msg = (
+                "Cannot replace parent node: "
+                f"{old_parent_label} is not a parent of {self.label}"
+            )
+            raise ValueError(msg)
```
willGraham01:
```suggestion
        super().replace_parent(old_parent_label, new_parent_label)
```

tests/test_algorithms/test_replace_node.py
willGraham01: There isn't anything special about Ricardo's graph with regards to this method, so I think you just want to write a bunch of parametrised tests on a smaller graph that will cover all the expected functionality. If those work then we can be confident it'll work with Ricardo's graph (though given we expect to be doing this exact transform on Ricardo's graph, this test may be worth keeping as a separate use-case).

Also, consider writing unit tests for `remove_edge` and `remove_node` (and `replace_parent` if you really want to make sure it's correctly implemented on each `Node` subclass).

Then testing `replace_node` can be done by creating a simple graph with like 4 nodes or something to use as a fixture, and checking various use-cases of the replacement:

- Doing a like-for-like replacement (EG just a name change or something)
- Reversing a selection of the edges connecting to the target node
- Isolating the node
- Etc

where in each case you confirm that the resulting copy-graph has the edge table you expect.

```python
@pytest.fixture
def graph_for_replacement_testing() -> Graph:
    ...

@pytest.mark.parametrise(...)
def test_replace_node(
    graph_for_replacement_testing,
    node_to_replace,
    new_node,
    expected_edge_table_after_replacement,
) -> None:
    
    g = replace_node(graph_for_replacement_testing, node_to_replace, new_node)

    # Psuedocode checks but you get what I mean
    assert node_to_replace not in g
    assert new_node in g
    # For sanity, consider giving the expected edge table in the form
    # (source_node_label, target_node_label) so you can give the input
    # as a list of tuples of strings. Then compare to node.label when you
    # fetch the actual edge table from g
    assert g.edge_table == expected_edge_table_after_replacement

```

willGraham01 (changes requested):
No faults on the method logic, but I think the code could be a bit cleaner in places.

Also there are 3 methods coming in, in this PR, and only one of them gets a test. Maybe something to address?

src/causalprog/graph/node/base.py
```diff
@@ -184,3 +184,12 @@ def assert_is_valid_value(self, value: float | npt.NDArray[float]) -> None:
         if self.shape != (value.shape if hasattr(value, "shape") else ()):
             msg = f"Invalid value for node: {self.label}"
             raise ValueError(msg)
+
+    def replace_parent(self, old_parent_label: str, _new_parent_label: str) -> None:
```
mscroggs: Do we want to somehow mark this function as for internal use only? A user can make a big mess of the graph if they use this directly rather than the replace_node algorithm. eg: this function does not change the edges in the graph so can lead to nodes having parents that the graph doesn't think are parents

willGraham01: We could just make it private (`_replace_parent`) but then ruff will complain at the `end.replace_parent` and `start.replace_parent` lines in `replace_node`.

So yeah I think best course of action is to put a sentence right at the top of the method's docstring, just to flag to people that they _really_ don't want to be calling this method directly.

tests/test_algorithms/test_replace_node.py
mscroggs: lots of unit tests of the different methods added in recent commits

tests/test_algorithms/test_replace_node.py
```diff
@@ -0,0 +1,131 @@
...
+        graph.add_node(ContinuousRandomVariableNode(label="A"))
+        graph.add_node(ContinuousRandomVariableNode(label="B", parents=["A"]))
+        graph.add_node(ContinuousRandomVariableNode(label="C"))
+        graph.add_node(ContinuousRandomVariableNode(label="D"))
+        graph.add_node(ContinuousRandomVariableNode(label="E", parents=["A", "B", "C"]))
+        graph.add_node(ContinuousRandomVariableNode(label="F", parents=["B"]))
+        graph.add_node(ContinuousRandomVariableNode(label="G"))
+        return graph
+
+    return _inner
```
mscroggs: This is duplicated in two files. Remind me to sort that

src/causalprog/graph/graph.py
```diff
@@ -254,3 +305,7 @@ def model(self, **parameter_values: npt.ArrayLike) -> dict[str, npt.ArrayLike]:
                 )
 
         return node_record
+
+    def is_dag(self) -> bool:
+        """Check if this graph is a directed acyclic graph."""
+        return nx.is_directed_acyclic_graph(self._graph)
```
mscroggs, line 307: There's no unit test for this method but as it's a direct wrap of a networkx function, i think that's ok?

src/causalprog/algorithms/replace_node.py
```diff
@@ -0,0 +1,49 @@
...
+        node_label_to_replace: The label of the node to be replaced.
+        replacement_node: The new node to be inserted.
+        new_graph_label: The label of the new graph.
+
+    Returns:
+        A copy of the graph with the replacement made.
+
+    """
+    if node_label_to_replace in replacement_node.parents:
+        msg = "Node being replace cannot be parent of replacement node"
```
willGraham01:
```suggestion
        msg = "Node being replaced cannot be parent of replacement node"
```

src/causalprog/graph/graph.py
```diff
@@ -90,6 +122,25 @@ def add_edge(self, start_node: Node | str, end_node: Node | str) -> None:
...
+        """
+        if isinstance(start_node, str):
+            start_node = self.get_node(start_node)
+        if isinstance(end_node, str):
+            end_node = self.get_node(end_node)
+        try:
+            self._graph.remove_edge(start_node, end_node)
+        except nx.exception.NetworkXError as err:
+            msg = "Cannot remove edge that is not in graph"
+            raise ValueError(msg) from err
```
willGraham01: Do we need to except and throw here? Unless the networkx error message itself isn't super clear, can't we just do `self._graph.remove_edge(...)` and delegate the error throw?

src/causalprog/graph/graph.py
willGraham01: Yeah I think we can let you get away with this 😂

tests/fixtures/graph.py
```diff
@@ -126,3 +131,23 @@ def _inner(mean: float, cov2: float) -> dict[str, npt.ArrayLike]:
         return {"X": x, "UX": ux}
 
     return _inner
+
+
+@pytest.fixture
+def seven_node_graph() -> Callable[[], Graph]:
+    """Creates a graph with seven nodes and a mixture of parent relations.
+
+    This graph is used to test node replacement."""
```
willGraham01:
```suggestion
    """Creates a graph with seven nodes and a mixture of parent relations.

    This graph is used to test node replacement, and is sketched below. 
    Note that nodes `D` and `G` are intentionally isolated.
    
    ```
    C <-- A --> B --> F    D    G
    |     |     |
    |     v     |
    ----> E <----
    ```
    """
```

willGraham01 (approved):
Though I found a typo and drew some ASCII art. Plus think we can get away with delegating one of the errors to `networkx` just so we don't have a nasty `try`-`catch` block in our codebase.

src/causalprog/graph/graph.py
mscroggs: I was trying to keep the networkx dependency limited to within one file just in case we ever want to change the graph library... but that's because of a conversation I had about UFL not causalprog, so in this case, yet let's just have networkx's error

PR #152, opened 2026-07-03
add mlps

samjmolyneux:


src/causalprog/mlps/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""MLPs."""
+
+from causalprog.mlps.mlp import FunctionalMLP, mlp
+
+__all__ = ["FunctionalMLP", "mlp"]
```
willGraham01:
```suggestion
__all__ = ("FunctionalMLP", "mlp")
```

Just convention, `__all__` is a tuple because we don't want it changed.

src/causalprog/mlps/mlp.py
```diff
@@ -0,0 +1,351 @@
...
+    def __call__(
+        self,
+        x: jax.Array,
+        *,
+        rngs: nnx.Rngs | None = None,
+    ) -> jax.Array:
+        x = self.linear(x)
+
+        if self.norm is not None:
+            x = self.norm(x)
```
willGraham01: Is there some way to avoid having to check for `None` here?

I'd much rather read

```python
    x = self.norm(x)
```

and trust that `self.norm` has been set to something like the identity map in the event that `None` was passed to `norm` in `__init__`.


Though I guess this might require us to write our own `flax`-compatible norm just for the identity map... so maybe it's not worth it. Same thing I guess applies to `self.dropout` too.

src/causalprog/mlps/mlp.py
```diff
@@ -0,0 +1,351 @@
...
+            return nnx.RMSNorm(num_features, rngs=rngs)
+        case None:
+            return None
+        case _:
+            msg = f"Unknown norm: {name}"
+            raise ValueError(msg)
+
+
+class _MLPBlock(nnx.Module):
+    """One MLP block."""
```
willGraham01, line 20: My non-ML brain is just trying to put together the pieces here; one MLP block is the thing that tells us how to "pass through" one layer of neurons? Or something different?

src/causalprog/mlps/mlp.py
```diff
@@ -0,0 +1,351 @@
...
+    match name:
+        case "layernorm":
+            return nnx.LayerNorm(num_features, rngs=rngs)
+        case "rmsnorm":
+            return nnx.RMSNorm(num_features, rngs=rngs)
+        case None:
+            return None
+        case _:
+            msg = f"Unknown norm: {name}"
+            raise ValueError(msg)
```
willGraham01: Small style comment but generally we try to split out self-contained functionality from files to prevent them from getting too long.

So here I'd move these two functions into a file called something like `mlps/_helpers.py` or `mlps/_specifiers.py` and then import them here. The leading underscore is because then the modules they define will be considered hidden modules by Python (similar to how they're hidden functions right now).

It also means that if you did want to test these functions, the split-up of the test files is quite natural too.

src/causalprog/mlps/mlp.py
```diff
@@ -0,0 +1,351 @@
...
+
+    graphdef, initial_parameters = nnx.split(model, nnx.Param)
+
+    return (
+        FunctionalMLP(
+            graphdef=graphdef,
+            has_dropout=dropout_rate > 0.0,
+        ),
+        initial_parameters,
+    )
```
willGraham01, line 239: I had a short foray into the `flax` docs, but just wanted to confirm that your idea with this code is:

- Build a state-ful MLP that reflects what the user has asked for.
- Split apart it's parameters (`initial_parameters`) and the model structure itself (`graphdef`, which is state-independent).
- Re-build it as a state-independent object by attaching the model structure to an instance of the `FunctionalMLP` class.
- Return the `FunctionalMLP` object which is now acting as the function we want in #139.

( Genius 👌)

src/causalprog/mlps/mlp.py
```diff
@@ -0,0 +1,351 @@
...
+        msg = "input_dim must be positive."
+        raise ValueError(msg)
+
+    if output_dim <= 0:
+        msg_0 = "output_dim must be positive."
+        raise ValueError(msg_0)
+
+    if dropout_rate < 0.0 or dropout_rate >= 1.0:
+        msg_2 = "dropout_rate must be in [0, 1)."
+        raise ValueError(msg_2)
```
willGraham01: Similar to the `_norm` and `_activation`, I'd be inclined to move these into an `mlp/_validation.py` file or something, just to reduce the file length.

tests/test_mlps/test_mlps.py
```diff
@@ -0,0 +1,539 @@
+"""Tests for explicit-parameter MLP builders."""
```
willGraham01, line 1: If you follow my comments on the `mlps/mlp.py` file, you can break this file up into multiple files too (typically by module). Tests are good don't get me wrong! But 500-line long files scare me slightly 😱

tests/test_mlps/test_mlps.py
```diff
@@ -0,0 +1,539 @@
...
+        ("layernorm", nnx.LayerNorm),
+        ("rmsnorm", nnx.RMSNorm),
+    ],
+    ids=[
+        "no-norm",
+        "layernorm",
+        "rmsnorm",
+    ],
+)
+def test_mlp_norms(norm: str | None, expected_norm_type: type) -> None:
```
willGraham01:
```suggestion
def test_mlp_same_norm_across_blocks(norm: str | None, expected_norm_type: type) -> None:
```

I think this is a bit more descriptive of you're actually testing right? 

(You are also checking that the `norm` strings map to the correct norm-function, but I think we can get away with checking that since you're handling that association with a `match` statement!)

tests/test_mlps/test_mlps.py
```diff
@@ -0,0 +1,539 @@
...
+    ],
+    ids=[
+        "relu",
+        "gelu",
+        "silu",
+        "tanh",
+        "identity",
+    ],
+)
+def test_mlp_activations(
```
willGraham01:
```suggestion
def test_mlp_same_activations_across_blocks(
```

Similarly reasons for suggesting along with my comment on the `norm` test above.

tests/test_mlps/test_mlps.py
```diff
@@ -0,0 +1,539 @@
...
+
+
+def test_mlp_different_seeds_give_different_initialisations(x_3: jax.Array) -> None:
+    f_0, theta_0 = build_mlp(seed=0)
+    f_1, theta_1 = build_mlp(seed=1)
+
+    y_0 = f_0(x_3, theta_0)
+    y_1 = f_1(x_3, theta_1)
+
+    assert not bool(jnp.allclose(y_0, y_1))
```
willGraham01: These can be combined into a single test via parametrisation


```suggestion
@pytest.mark.parametrize("use_same_seed", [True, False])
def test_mlp_seed_and_init(
    x_3: jax.Array, seed_0: int = 0, *, use_same_seed: bool
) -> None:
    seed_1 = seed_0 if use_same_seed else seed_0 + 1

    f_0, theta_0 = build_mlp(seed=seed_0)
    f_1, theta_1 = build_mlp(seed=seed_1)

    y_0 = f_0(x_3, theta_0)
    y_1 = f_1(x_3, theta_1)

    assert jnp.allclose(y_0, y_1) == use_same_seed
```

tests/test_mlps/test_mlps.py
```diff
@@ -0,0 +1,539 @@
...
+    rngs_0 = nnx.Rngs(params=123)
+    rngs_1 = nnx.Rngs(params=123)
+
+    f_0, theta_0 = build_mlp(rngs=rngs_0)
+    f_1, theta_1 = build_mlp(rngs=rngs_1)
+
+    y_0 = f_0(x_3, theta_0)
+    y_1 = f_1(x_3, theta_1)
+
+    assert bool(jnp.allclose(y_0, y_1))
```
willGraham01: I think we can leave out this test case. Not that it was bad idea to check it - but it ultimately boils down to just a check that `nnx` is assembling it's random states correctly under the hood (and given that we're using `flax`, we should take such things as trusted).

tests/test_mlps/test_mlps.py
```diff
@@ -0,0 +1,539 @@
...
+        "hidden-dims-with-hidden-layers-and-units",
+        "hidden-dims-contains-zero",
+        "missing-hidden-configuration",
+        "missing-hidden-layers-with-hidden-units",
+        "negative-hidden-layers",
+        "positive-hidden-layers-missing-units",
+        "positive-hidden-layers-zero-units",
+    ],
+)
+def test_mlp_raises_on_invalid_deep_layers(
```
willGraham01: Couple of points here:

We have a fixture called `raises_context` that wraps around `pytest.raises` to make parametrisations like this a bit nicer looking.

Rather than

```python
def test_mlp_raises_on_invalid_deep_layers(
    ...,
    message: str,
):
    with pytest.raises(ValueError, match=message):
        ...
```

you can use

```python
def test_mlp_raises_on_invalid_deep_layers(
    ...,
    expected_error: Exception,
    raises_context,
):
    with raises_context(expected_error):
        ...
```

You will need to change your parametrisation mark to be over "expected_error" instead of "message" too.

Second point - this is actually just a test of `_resolve_hidden_dims` right? If so, just call that function directly rather than the whole `build_mlp` 😄 If you've split `resolve_hidden_dims` out into another file, you can do the same thing for this test into a separate test file.

tests/test_mlps/test_mlps.py
```diff
@@ -0,0 +1,539 @@
...
+        (3, 2, 1.0, "dropout_rate must be in"),
+    ],
+    ids=[
+        "zero-input-dim",
+        "zero-output-dim",
+        "negative-dropout-rate",
+        "dropout-rate-one",
+    ],
+)
+def test_mlp_rejects_invalid_configurations(
```
willGraham01: Similar to my comment above - you can use `raises_context` here, and you are effectively just calling `_validate_mlp_base_config`, so might as well do that directly.

tests/test_mlps/test_mlps.py
```diff
@@ -0,0 +1,539 @@
...
+
+
+def test_mlp_rejects_unknown_activation() -> None:
+    with pytest.raises(ValueError, match="Unknown activation"):
+        build_mlp(activation="not_an_activation")
+
+
+def test_mlp_rejects_unknown_norm() -> None:
+    with pytest.raises(ValueError, match="Unknown norm"):
+        build_mlp(norm="batchnorm")
```
willGraham01: All three of these are checks against `_norm` or `_activation` behaviour (and can use the `raises_context` fixture). Like with my comments about `_validate_mlp_base_config` and `resolve_hidden_dims`, the these inner functions directly to check for the error (and break out into other test files if needed).

tests/test_mlps/test_mlps.py
```diff
@@ -0,0 +1,539 @@
...
+        f,
+        theta,
+        x_train,
+        y_train,
+        learning_rate=0.01,
+        steps=1_000,
+    )
+
+    assert float(final_loss) < float(initial_loss)
+    assert float(final_loss) < 1e-5
```
willGraham01:
```suggestion
@pytest.mark.parametrise(
    ("y_func", "mlp_kwargs", "fitting_kwargs"),
    [
        pytest.param(
            lambda x: 2.0 * x + 0.5,
            {
                "input_dim": 1,
                "output_dim": 1,
                "hidden_dims": [],
                "activation": "identity",
                "norm": None,
                "dropout_rate": 0.0,
            },
            {
                "learning_rate": 0.05,
                "steps": 300,
            },
            id = "Learn linear problem",
        ),
        pytest.param(
            lambda x: x**3 + 0.3 * x - 0.2,
            {
                "input_dim": 1,
                "output_dim": 1,
                "hidden_layers": 2,
                "hidden_units": 16,
                "activation": "relu",
                "norm": None,
                "dropout_rate": 0.0,
            },
            {
                "learning_rate": 0.01,
                "steps": 1_000,
            },
            id = "Learn nonlinear problem",
        ),
    ]
)
def test_mlp_learning(
    y_func: Callable,
    mlp_kwargs,
    fitting_kwargs,
    seed: int,
    tolerance: float = 1e-5,
) -> None:
    x_train = jnp.linspace(-1.0, 1.0, 128).reshape(-1, 1)
    y_train = y_func(x_train)

    f, theta = mlp(
        **mlp_kwargs,
        seed=seed,
    )

    _, initial_loss, final_loss = fit_mlp_to_targets(
        f,
        theta,
        x_train,
        y_train,
        **fitting_kwargs,
    )

    assert float(final_loss) < float(initial_loss)
    assert float(final_loss) < tolerance
```

I can squash these two into a single parametrised test (so if we do want to add extra examples, we only need to add another parametrisation).

(Also `seed` is one of our global fixtures, like `raises_context`. It is still just 0, but inserting it here for consistency).

willGraham01 (changes requested):
I think this is a good piece of work and nothing else needs adding, just think it could do with some re-organisation to break up the larger files into several smaller ones 😄 If we spot anything that we do want to add on top of what's here, we'll put it in an issue for later.

I've put some suggestions in specific places. I haven't quite finished looking through the tests, but these should be a good first pass to see where we end up. Broadly speaking:

- Let's try and break up the tests across several files, based on what they're actually testing.
- Where possible, we should call the "smallest" (or "innermost") function possible that we're actually checking the behaviour of. See my comments on the test file relating to calling `_resolve_hidden_dims`, `_norm`, over `build_mlp` for example.
- We have two fixtures: `seed` and `raises_context`, that can be used in a few of these tests.

Once those changes are made, I'll do another pass and we can see how things look.

src/causalprog/mlps/mlp.py
samjmolyneux: Yes, exactly!

src/causalprog/mlps/mlp.py
samjmolyneux: Yes, it is their documented way of doing it. I can't take much credit 😅 

I had mentioned that I was concerned it might negatively affect performance, but apparently it's faster than the default stateful method: [flax performance considerations](https://flax.readthedocs.io/en/stable/guides/performance.html#functional-training-loop)

src/causalprog/mlps/mlp.py
samjmolyneux: Come to think of it, I think people often separate out the entire block into its own file with the block components. So maybe `_mlp_block.py` with `_activation(...)`, `_norm(...)` and `_MLPBlock`? 

Or do you prefer just `_activation(...)` and `_norm(...)` in a separate one?

tests/test_mlps/test_mlps.py
samjmolyneux: The 2nd and 3rd do yes, so will move them into `test_validators.py`. 

1st was testing the original FunctionalMLP to make sure dropout was being added in the correct conditions, but I have changed it so that flax handles that now. So I'm removing the first test.

src/causalprog/mlps/mlp.py
samjmolyneux: I ended up justing putting the specifiers in a separate file as we discussed in our friday meeting.

samjmolyneux (commented):
I've pretty much made all the changes you recommended.

I also learnt that flax naturally handles all the dropout stuff that I was checking for in the original pr, so no need to include it or remove it or change the api for calling StatefulMLP based on wether it is included. 

The changes that you might want me to still make are to `test_mlp_uses_correct_hidden_configuration`, `test_mlp_same_norm_accross_blocks` and `test_mlp_same_activations_across_block` because they are currently testing that both the specifier works correctly and that the mlp then implements them correctly across the blocks. 

If there's anything I've missed please let me know!

tests/test_mlps/test_validators.py
```diff
@@ -0,0 +1,158 @@
...
+        "negative-dropout-rate",
+        "dropout-rate-one",
+    ],
+)
+def test_validate_mlp_base_config_rejects_invalid_configuration(
+    input_dim: int,
+    output_dim: int,
+    dropout_rate: float,
+    expected_error: Exception,
+    raises_context: Callable[[Exception], AbstractContextManager[object]],
```
willGraham01:
```suggestion
    raises_context,
```

Just so you don't have to write out this typehint everywhere, Matt & I have disabled ruff's "must have argument typehints" for test functions. I've added a suggestion to delete the typehint where-ever I've spotted it being used, but if you want to keep it feel free.

So for fixtures like `raises_context`, you can just leave it without a typehint when you use it as an argument to a test function 😄

src/causalprog/mlps/mlp.py
```diff
@@ -0,0 +1,239 @@
+"""MLP function builder."""
```
willGraham01:
```suggestion
"""Multi-Layer Perceptron (MLP) function builder."""
```

Because TLAs should always be expanded once.

tests/test_mlps/conftest.py
```diff
@@ -0,0 +1,10 @@
+"""MLP test fixtures."""
+
+import jax
+import jax.numpy as jnp
+import pytest
+
+
+@pytest.fixture
+def x_3() -> jax.Array:
+    return jnp.array([1.0, -2.0, 0.5])
```
willGraham01: As the `x_3` fixture is only used in `test_mlps.py`, I'd leave it as a fixture defined in that file. (`conftest` lets us share fixtures between files, but right now this guy doesn't need to be shared).

tests/test_mlps/test_mlps.py
```diff
@@ -0,0 +1,429 @@
...
+def test_mlp_forward_pass_calls_layers_in_expected_order(
+    x_3: jax.Array,
+    seed: int,
+) -> None:
+    def record_call(calls, name, fn):
+        def wrapped(*args, **kwargs):
+            calls.append(name)
+            return fn(*args, **kwargs)
+
+        return wrapped
```
willGraham01, line 301: I quite like this function and how you use it to make the MLP model "noisy", for lack of a better word 😂 

I think there is going to be some utility in refactoring it out, as I can imagine there will be other cases (particularly integration tests) where we want to do something similar.

But right now, it's not critical we do this in this PR. #157

tests/test_mlps/_helpers.py
```diff
@@ -0,0 +1,69 @@
+"""Helpers for MLP tests."""
```
willGraham01, line 1: We should elevate these to global fixtures, in a similar way we do with `raises_context` and a few others. 

I've tracked this in #157, as again there's no reason to do this right now. But I imagine we'll want to use `build_mlp` (in particular!) in a lot of the tests going forwards.

tests/test_mlps/test_specifiers.py
```diff
@@ -0,0 +1,34 @@
...
+from causalprog.mlps._specifiers import (
+    ActivationName,
+    NormName,
+    resolve_activation,
+    resolve_norm,
+)
+
+
+def test_resolve_activation_rejects_unknown_activation(
+    raises_context: Callable[[Exception], AbstractContextManager[object]],
```
willGraham01:
```suggestion
    raises_context,
```

tests/test_mlps/test_specifiers.py
```diff
@@ -0,0 +1,34 @@
...
+) -> None:
+    unknown_activation = cast("ActivationName", "not_an_activation")
+
+    with raises_context(ValueError("Unknown activation: not_an_activation")):
+        resolve_activation(unknown_activation)
+
+
+def test_resolve_norm_rejects_unknown_norm(
+    raises_context: Callable[[Exception], AbstractContextManager[object]],
```
willGraham01:
```suggestion
    raises_context,
```

tests/test_mlps/test_validators.py
```diff
@@ -0,0 +1,158 @@
...
+        "positive-hidden-layers-missing-units",
+        "positive-hidden-layers-zero-units",
+    ],
+)
+def test_resolve_hidden_dims_rejects_invalid_configuration(
+    hidden_dims: Sequence[int] | None,
+    hidden_layers: int | None,
+    hidden_units: int | None,
+    expected_error: Exception,
+    raises_context: Callable[[Exception], AbstractContextManager[object]],
```
willGraham01:
```suggestion
    raises_context,
```

willGraham01 (approved):
Great stuff, thanks Sam! There's a couple of minor points down below that I've added, but am approving because nothing else is blocking.

There are a couple of things we could do to make the helper-functions accessible to the rest of the test suite, which I've opened #157 for. No need to address those points in this PR though - let's use separate ones for those.

Issue #133, opened 2026-06-04
Different forms of the loss function B (in terms of r and $\hat{r}$)


mscroggs:
Make learn initaliser method (in #131) take Python function for B as an input.

B_builder:
   - inputs: r, $\hat{r}$
   - returns: callable B(theta) -> float

The learn initialiser should return the callable B that B_builder generated alongside $\theta^*$

PR #160, opened 2026-07-14
Learn Initialiser Method

### LINKED ISSUE #131: Learn initialiser method
Again, hard code Ricardo's case (see #130).

- Inputs: graph, $\hat{r}$ as a Python callable (user has done pre training to get this), $\hat{\theta}_x$, QuadratureMethod (#128), set of evaluation points
- Returns: $\theta^*$, the argmin of $B(\theta):=\tfrac1n\sum_i(\hat{r}_i-r_i)^2$ where $_i$ denotes at the $i$ th evaluation point

For now, use gradient descent method for the optimisation

willGraham01:
Resolves #131 |

Adds the `learn_initialiser` method to the `graph/ricardo.py` module. `learn_initialiser` determines the argmin of the function $B(\theta)$, via what is essentially least-squares residual minimization over a set of evaluation points.

Some things were a little bit fiddly to implement. Namely:

- `evaluation_points` is a dictionary of arrays, rather than just a simple array, due to how we are creating the graph. As such, the docstring explains how to format `evaluation_points` in such a way that `learn_initialiser` can efficiently map over them. This also necessitates the presence of the `evaluation_points_axes_mapping` argument, which is needed when data for particular nodes is a mixture of scalar / vector / matrix values.
- Though #131 states to just use gradient descent for the optimization right now, the implementation in this method defaults to using our `stochastic_gradient_descent` to be consistent (and because we'll want to be doing stochastic descent in general, anyway).
- `learn_initialiser` has keyword arguments that are passed to the solver being used, for full flexibility. This technically means that any solver the user wants to use can be configured via this method.
- We don't support alternative forms of the function $B$, but this is a stretch goal #133.

In terms of testing; I came up with the following few cases:

- Did some testing for the various `evaluation_points_axes_mapping` configurations, to check that this was respected.
- Wrote some simple tests that check we can optimise a deterministic function, with reasonable success.
- Wrote some simple tests for a built regression function (using `build_regression_function`) in cases that were simple enough that I could determine the answers analytically.

We probably want a full integration test that checks the general case, IE when `theta` is of the form to be expected for MLPs, though.


src/causalprog/graph/ricardo.py
```diff
@@ -187,3 +190,104 @@ def _r(
...
+    across its first argument. This means that `evaluation_points` ($\mathcal{D}$)
+    should be passed in a suitable format for `jax.vmap` to map over. For all-scalar
+    nodes, this would simply be a dictionary whose keys are 1D arrays of the same shape
+    as `r_hat_i`. "Slices across the keys" of this dictionary correspond to individual
+    evaluation points $i$; for example passing
+    `evaluation_points = {"x": [0, 1], "z": [10, 20]}` corresponds to
+    $mathcal{D} = \{ (0, 10), (1, 20) \}$. When mixing scalar- and vector-valued nodes,
+    use `evaluation_points_axes_mapping` to specify which axes of each key-value
+    corresponds to the axes over which to vectorise the inputs (default is axis 0).
+    For example,
```
samjmolyneux: Should 'keys' be 'values' in this paragraph?

src/causalprog/graph/ricardo.py
```diff
@@ -187,3 +190,104 @@ def _r(
...
+    - $\hat{r}(x, z, l)$ is a learnt estimate of the regression function,
+    - $r(x, z, l; \theta)$ is the estimate of the regression function using the graph
+    structure,
+    - $\theta$ are the model parameters over which to optimise,
+    - and the summation is taken over a set of evaluation points
+    $\mathcal{D} = \left\{ (x^{(i)}, z^{(i)}, l^{(i)}) \right\}_{i=1}^N$.
+    Subscript $i$s denote evaluation at the $i$-th evaluation point.
+
+    To evaluate $r_i$, the function will attempt to vectorise the input function `r`
+    across its first argument. This means that `evaluation_points` ($\mathcal{D}$)
```
samjmolyneux:
```suggestion
To evaluate $r_i$, `learn_initialiser` will attempt to vectorise `r` across `r`'s first argument. This means that `evaluation_points` ($\mathcal{D}$)
```

Up to if you prefer original and dislike verbosity, but I get confused easily by which function is which.

src/causalprog/graph/ricardo.py
```diff
@@ -187,3 +190,104 @@ def _r(
...
+
+    corresponds to $\mathcal{D} = \{((0, 3, 6), 10), ((1, 4, 7), 20), (2, 5, 8), 30)\}$.
+    It is only necessary to specify arrays that are not mapping over their `0`th axes in
+    `evaluation_points_axes_mapping`.
+
+    Args:
+        r: Regression function, $r$. Typically the output of `build_regression_function`
+        evaluation_points: Set of evaluation points, $\mathcal{D}$
+        r_hat_i: The values of the estimate of r at the evaluation points, $\hat{r}_i$
+        evaluation_points_axes_mapping: Axes to vectorising over when evaluating $r$
```
samjmolyneux:
```suggestion
        evaluation_points_axes_mapping: Axes to vectorise over when evaluating $r$
```

src/causalprog/graph/ricardo.py
```diff
@@ -187,3 +190,104 @@ def _r(
...
+    )
+    _r = jax.vmap(r, in_axes=in_axes)
+    n_eval = r_hat_i.shape[0]
+
+    def _objective_function(theta: ModelParam) -> ModelParam:
+        r"""Evaluate $B(\theta)$."""
+        r = _r(evaluation_points, theta)
+        return ((r_hat_i - r) ** 2).sum() / n_eval
+
+    return optimiser(_objective_function, *optimiser_args, **optimiser_kwargs)
```
samjmolyneux: Might be a little confusing how we have a high-level optimiser such as `stochastic_gradient_descent` that runs the entire minimisation, and low-level optimisers such as the `optax` ones that only perform a single step.  

So the optimiser `stochastic_gradient_descent` takes an optimiser arg of type `optax.GradientTransformationExtraArgs`.

Would be good to separate them, such as calling the low-level optimiser 'optimiser' and high-level one 'solver'

src/causalprog/graph/ricardo.py
```diff
@@ -187,3 +190,104 @@ def _r(
...
+
+
+def learn_initialiser(
+    r: MLPAlias,
+    evaluation_points: dict[str, NDArray],
+    r_hat_i: NDArray,
+    *,
+    evaluation_points_axes_mapping: dict | None = None,
+    optimiser: Callable | None = None,
+    optimiser_args: tuple = (),
```
samjmolyneux: Just wondering if we can do something about the initial guess? 

Currently we have `initial_guess` as a required arg to `stochastic_gradient_descent`, but for this function we have have to include it in the optional `optimiser_args` kwarg here. If we leave it empty then things will break. 

Wondering if the best solution is to have `initial_guess` be optional for `stochastic_gradient_descent` and have `stochastic_gradient_descent` pick a random start point when `initial_guess` is left empty. 

Anyway, not important functionality-wise so not important to deal with for the time being.

src/causalprog/graph/ricardo.py
```diff
@@ -187,3 +190,104 @@ def _r(
...
+        data_axes,
+        None,
+    )
+    _r = jax.vmap(r, in_axes=in_axes)
+    n_eval = r_hat_i.shape[0]
+
+    def _objective_function(theta: ModelParam) -> ModelParam:
+        r"""Evaluate $B(\theta)$."""
+        r = _r(evaluation_points, theta)
+        return ((r_hat_i - r) ** 2).sum() / n_eval
```
samjmolyneux: I haven't thought about this hard enough yet, but I'm wondering how likely we are to get subtle broadcasting errors here, given that r_hat_i is user supplied. 

Maybe we can somehow do something like 

```python
r_values=_r(evaluation_points, initial_guess)

if r_values.shape != r_hat_i.shape:
    raise ValueError(
        "`r_hat_i` and the vectorised regression output must have "
        "matching shapes, but got "
        f"{r_hat_i.shape} and {r_values.shape}."
    )
```
before running the optimiser. 

But I'm not sure, my brain is very tired. I will recheck this section tomorrow.

src/causalprog/graph/ricardo.py
```diff
@@ -187,3 +190,104 @@ def _r(
...
+    data_axes = dict.fromkeys(evaluation_points, 0)
+    data_axes.update(evaluation_points_axes_mapping)
+    in_axes = (
+        data_axes,
+        None,
+    )
+    _r = jax.vmap(r, in_axes=in_axes)
+    n_eval = r_hat_i.shape[0]
+
+    def _objective_function(theta: ModelParam) -> ModelParam:
```
samjmolyneux:
```suggestion
    def _objective_function(theta: ModelParam) -> NDArray:
```
or 
```suggestion
    def _objective_function(theta: ModelParam) -> jax.Array:
```
?

src/causalprog/graph/ricardo.py
samjmolyneux: Just thinking that we have `r`, which is our r function,
```python
r: MLPAlias,
```
`_r` which is our vectorised r function,
```python
_r = jax.vmap(r, in_axes=in_axes)
```
and then `r` again which is the vectorised r function parameterised with `theta`, 
```python
r = _r(evaluation_points, theta)
```

Could rename something like `r`, `vectorised_r`, `r_theta`? 
Or `r`, `r`, `r_theta`?

samjmolyneux (commented):
- Tests all look good to me, again took me longer than I'm proud to admit to wrap my head around some of the test cases 😅 . 

- I made a few suggestions for `learn_initialiser(...)`. Nothing major so feel free to reject any and all as you see fit. The only thing worth mentioning is a potential subtle error from broadcasting that could come about. Just thinking that r_hat_i could be produced by different models (e.g. XGBoost, RandomForest etc) that could use different dimensions that could cause broadcasting in `_objective_function(...)` when the shapes of `r_hat_i` and `r` don't match. I've not thought hard enough about it yet though so will have a think tomorrow.

src/causalprog/graph/ricardo.py
willGraham01: Yeah I wrangled with this myself for a little while. Originally I had `initial_guess` be a required argument to this function, but then realised that different `optimiser`s might take it as different positional arguments.

Have opened #164 to track this

src/causalprog/graph/ricardo.py
willGraham01: I've changed the inputs to `learn_initialiser` to reflect this - using `solver` now over `optimiser` for this reason.

src/causalprog/graph/ricardo.py
willGraham01: I think it's keys, though I'm not sure :sweat_smile: 

I guess what would you call this (hypothetical) operation:

```python
>>> my_dict = {
    "a": [0, 1, 2],
    "b": [3, 4, 5],
    "c": [6, 7, 8],
}

>>> # Hypothetical operation starts here
>>> my_dict[0]
{"a": 0, "b": 3, "c": 6}
>>> my_dict[1:2]
{"a": [1, 2], "b": [4, 5], "c": [7, 8]"}
```

I went with "slicing across the keys", though now that I write it out maybe "slicing across the values" is more descriptive of what's going on :sweat_smile:

src/causalprog/graph/ricardo.py
willGraham01: We should only ever have scalar-valued regression functions, so in theory `r_hat_i` should always be provided as a 1D array. Though you're right that if it's provided as a column vector rather than a row vector, we might encounter broadcasting problems.

My inelegant solution is to do a `jnp.squeeze` on `r_hat_i`, which should eliminate any misleading dimensions. Likewise, then the `jax.vmap`-ed $r$ will also output a 1D function, so we shouldn't hit broadcasting problems.

willGraham01:
I think all comments are now addressed, but feel free to push back on my solutions if you don't agree with how I've gone about them!

src/causalprog/graph/ricardo.py
samjmolyneux: Yeah for 'slicing across keys' I guess both kind of sound right, I don't know! 😅 

For this one though:
```
 whose keys are 1D arrays of the same shape as `r_hat_i`
 ```
 It's the values that are 1D arrays when our nodes are scalar valued right? Am I misunderstanding this?

Comment (willGraham01) [reply src/causalprog/graph/ricardo.py]: Lol yeah, that one is definitely values :joy:

**Review by samjmolyneux — APPROVED** [review]
Yeah this looks all good to me. 

I forgot that Y is scalar-valued only, so of course, as you show, we can just check the dimensions. 

I would probably favour removing 
```python 
r_hat_i = r_hat_i.squeeze()
```
now that we have the ndim check, and thus leaving the responsibility on the user to make sure the `r_hat_i` they enter has the correct shape. If it isn't they will quickly find out and sometimes `.squeeze` is a little unpredictable! 

Anyway, I'm happy to approve!

Issue #161, opened 2026-07-17
Flatten MLP inputs


willGraham01:
Our MLPs current expect array-like data inputs, however we plan to provide `PyTree`-like objects (mostly `dict`s) as their data inputs.

This means that we need to `flatten` the `dict` inputs that MLPs receive, so that they can then be propagated through the MLP to produce a prediction. For now, we'll do this on the MLP side, but in future we should look into alternative means of doing this.

PR #162, opened 2026-07-20
`FunctionalMLP` inputs can be `PyTree`s

willGraham01:
Addresses #161 |

Allows `FunctionalMLP`s to take inputs in the form of `PyTree` objects, rather than only 1D column vectors. This is achieved by unraveling the input object before passing to the underlying model that uses column vectors.

Key changes:

- The `mlp` builder function inside `causalprog.mlps.mlp` now takes `input_dim` as either an `int` or PyTree object.
  - The behaviour when it is passed as an `int` is identical to the behaviour prior to this PR, and this PR does not change this behaviour.
  - If passed as a `PyTree`, each leaf of `input_dim` must be a `jax.Array` of integers that specifies the shape of the array expected by that leaf for the input data. For example, if we want to pass input data in the format `{"x": [1, 2, 3], "y": 0, "z": [[1, 2], [3, 4]]}`, then `input_dim` should be passed as `{"x": 3, "y": 1, "z": jnp.array([2, 2])}`.
  - **NOTE** The requirement that shapes be passed as `jax.Array`s and not `tuple`s is because `tuple`s are not considered leaves of `PyTree`s.
- `FunctionalMLP` now has an explicitly implemented constructor method.
  - Previously frozen attributes have been made private to mimic the old behaviour, but allow for an `__init__` method that sets instance attributes derived from the constructor arguments.
  - The static methods `identity` and `unravel_tree` are convenient aliases for the conversion between `PyTree` inputs (that the user will provide) and 1D-array inputs (that the underlying `flax` models want to use). `identity` is present for the case when the user is genuinely providing input data as a 1D-array already.
  - The `data_format` property has been added to instances, so the user can view the format of the input data that should be expected. This has the same interpretation as the  `input_dim` argument to `mlp`, explained above.
  - The public `data_as_flat_array` method has been added. This method is the public-facing means of converting whatever input format the user provides into a 1D-array that `flax` can use. It aliases either `identity` or `unravel_tree`.
  - `__call__` for `FunctionalMLP`s will now cast the input data via `data_as_flat_array` before calling the underlying `flax` model, which is expecting a 1D array.
    - Note that whilst we implicitly supported batching prior to this PR, we now no longer can support batching on `PyTree`-like input data formats.
    - The exception to this is when the `FunctionalMLP` explicitly takes 1D-array-data directly, then `__call__` can still be used for batching. A note in the docstring has been added about this, but generally we want to discourage this. Though it is probably worth updating #161 or opening a follow-on issue to ensure this is documented somewhere (as we likely will want to allow for batching in the future).

Minor other changes that got caught up:

- Created a `causalprog._types` module since I was needing to import the `PyTree` type alias, which appears to not be consistently imported from one place within the package. This is now rectified, the `PyTree` typevar is now always imported from one place.


src/causalprog/_types.py
```diff
@@ -0,0 +1,6 @@
+from typing import TypeAlias, TypeVar
+
+from jaxlib._jax.pytree import PyTreeDef as PyTreeDefNative
```
samjmolyneux:
```suggestion
from jax.tree_util import PyTreeDef as PyTreeDefNative
```
Looks like they have a public version. It exposes the exact same class, just through a public module.

src/causalprog/mlps/_validation.py
```diff
@@ -1,5 +1,33 @@
...
+    the event that the input is a `PyTree`, the leaves of the `PyTree` are
+    interpreted as the shapes of the arrays at the corresponding leaves.
+
+    Args:
+        input_data_format: `PyTree` whose leaves contain shape information.
+
+    Returns:
+        Total number of elements in the `PyTree`, based on shape information.
+
+    """
```
samjmolyneux, line 37:
```suggestion
    """
    Calculate the total number of input elements described by a shape PyTree.

    This function is primarily used to determine the flattened input dimension
    of a `FunctionalMLP` from its input-data format specification.

    Each leaf of `input_data_format` describes the shape of the corresponding
    input-data leaf:

    - A scalar integer `n` represents a one-dimensional array containing `n`
      elements.
    - A one-dimensional JAX array such as `jnp.array([2, 3])` represents an
      array with shape `(2, 3)` and therefore contains `2 * 3 = 6` elements.

    Multidimensional shapes must be represented as JAX arrays rather than Python
    tuples, such as the `(2, 3)` returned by `jnp.zeros((2, 3)).shape`. Tuples
    are treated by JAX as PyTree containers, so `(2, 3)` would be interpreted
    as two separate scalar leaves rather than one shape specification.

    Args:
        input_data_format: An integer or PyTree whose leaves contain shape
            specifications.

    Returns:
        The total number of input elements represented by all shape leaves.

    """
```

I feel like this info should be somewhere, whether it should be here or in the `mlp(...)` func I'm not sure.
 
But despite the fact that you make it obviously clear that the shape info should be a `jax.Array`, my brain still instinctively just thought I just have to do `arr.shape`, which of course would be wrong and then I found myself confused. I think it would be useful to directly address it somewhere, similar to how I did above.

src/causalprog/mlps/mlp.py
```diff
@@ -90,15 +95,51 @@ def __call__(
...
+        """
+        return self._data_format
+
+    @property
+    def graphdef(self) -> nnx.GraphDef:
+        """Model graph definition, for view access only."""
+        return self._graphdef
+
+    @staticmethod
+    def identity(data: PyTree) -> jax.Array:
```
samjmolyneux:
```suggestion
    def identity(data: jax.Array) -> jax.Array:
```

samjmolyneux (approved):
Looks good to me, just a couple of minor suggestions that I may be wrong on. 

Only real suggestion is just to maybe add a little more context around the `input_data_format`. For some reason my brain just thought `.shape()` would be okay, but obviously it isn't. Can't help but imagine others would make the same mistake then, so I just suggested spoon feeding that info to the user (details in the suggestion). 

Also, entirely up to you, but I would be tempted to add a test somewhere that the mlp outputs the same output for the array and the PyTree. 
```python
def test_mlp_pytree_input_matches_merged_model() -> None:
    """Check that FunctionalMLP correctly flattens PyTree inputs before evaluation."""
    input_fmt = {
        "a": 1,
        "b": jnp.array([2]),
        "c": jnp.array([2, 2]),
    }
    pytree_input = {
        "a": jnp.array(0.5),
        "b": jnp.array([1.0, -2.0]),
        "c": jnp.arange(4.0).reshape(2, 2),
    }

    f, theta = build_mlp(input_dim=input_fmt)
    merged_model = nnx.merge(f.graphdef, theta)

    actual = f(pytree_input, theta)
    expected = merged_model(f.data_as_flat_array(pytree_input))

    assert actual.shape == expected.shape
    assert bool(jnp.allclose(actual, expected))
```

Issue #158, opened 2026-07-13
Moving `_helpers.py` into the `conftest` scope


willGraham01:
Currently the `mlp` tests import from a local helper file, which is fine since these functions are not needed anywhere else (right now). `build_mlp` (and a few others) might be wanted elsewhere in the near future however, and relative imports aren't super robust when we start moving all around the test suite. We should elevate the functions in `test_mlps/_helpers.py` to actual fixtures, in our `tests/fixtures` scope.

- Create a new file, `tests/fixtures/mlps.py`
- Add the functions `build_mlp` and `fit_mlp_to_targets` as `pytest.fixtures`

Note that since we want to make _functions_ fixtures, we need to define them as fixtures that return functions. For `build_mlp` for example,

```python
@pytest.fixture
def build_mlp() -> Callable[..., tuple[FunctionalMLP, nnx.State]]:
    # default_kwargs could even be it's own fixture, that `build_mlp` depends on.
    default_kwargs = {
        "input_dim": 3,
        "output_dim": 2,
        "hidden_layers": 3,
        "hidden_units": 8,
    }

    def _build_mlp(**overrides) -> tuple[FunctionalMLP, nnx.State]:
        # Dicts are mutable and persistent, hence why we're wrapping
        # default_kwargs in a dict constructor here (to make a copy before updating)
        kwargs = dict(default_kwargs).update(overrides)
        return mlp(**kwargs)

    return _build_mlp
```

Then instead of

```python
from ._helpers import build_mlp

def my_test(...):
    build_mlp(...)
```

we'd do

```python
def my_test(..., build_mlp):
    build_mlp(...)
```

A similar thing will need to be done for `fit_mlp_to_targets`.

PR #163, opened 2026-07-20
Causal response function for Ricardo graph

### LINKED ISSUE #134: Function for evaluating d
This will be similar to the $r$ -builder in #129; we'll need a function `d_builder` (or `d_maker`) that takes in a graph $G$ representing Ricardo's graph. Using this information, it needs to evaluate the function $d(x, l; \theta)$ as defined in Ricardo's document. We're hard-coding the functional form of this method, for now.

The function $d(x, l; \theta)$ is defined as

$$
d(x, l; \theta) = \mathbb{E}[Y \vert \text{do}(X=x), L=l]
= \int f_Y(u_y, x, l; \theta_Y)p_N(u_y) \mathrm{d} u_y
$$

where $p_N$ is the PDF of a normal distribution with mean 0 and standard deviation 1.

We need to write a function

```python
def d_builder(graph, n_points) -> Callable[[dict, dict], float]:
```

that takes Ricardo's graph `graph` as an input. Given the form of the integral it makes the most sense for us to start by hard-coding in use of the `causalprog.quadrature.UniformWeightMonteCarloGaussianQuadrature` method, and hence the `n_points` input to `d_builder` should simply be used to create an instance of this quadrature class, for evaluating the integral. Using `UniformWeightMonteCarloGaussianQuadrature.integrate` with an integrand of $f_Y$ should then be sufficient to evaluate $d$ (see the class docstring for more information).

`d_builder` needs to return a function `d` with signature `d(u_yxl, theta)`. `u_yxl` should be a dictionary with keys `"u_y"`, `"x"`, `"l"` that correspond to the argument values $u_y$, $x$, $l$ that $d$ takes. `theta` should be a dictionary with keys such as `"theta_y"`, `theta_x`, etc, and corresponds to the $\theta$ argument to $d$.

samjmolyneux:
Closes #134 

* Separated `vectorise_over_dict_args` out so it can be reused in the tests.
* Added `build_causal_response_function(...)` to `src/causalprog/graph/ricardo.py`.
* Added three tests to `tests/test_integration/test_ricardo_causal_response_function.py`:

  1. `test_fy_independent_of_uy(...)` uses the fact that if $`f_Y`$ is independent of $`u_y`$, then

     $`d(x,l;\theta)=\int_{-\infty}^{\infty}f_Y(u_y,x,l;\theta_Y)\,p_{\mathcal{N}}(u_y;0,1)\,\mathrm{d}u_y`$<br>
     $`=f_Y(x,l;\theta_Y)\int_{-\infty}^{\infty}p_{\mathcal{N}}(u_y;0,1)\,\mathrm{d}u_y`$<br>
     $`=f_Y(x,l;\theta_Y).`$

  2. `test_causal_response_matches_standard_normal_moments(...)` uses a quadratic response function of the form

     $`f_Y(u_y,x,l;\theta_Y)=\theta_{\mathrm{quadratic}}u_y^2+\theta_{\mathrm{linear}}u_y+\theta_xx+\theta_l\sum_i l_i.`$

     Since $`U_Y\sim\mathcal{N}(0,1)`$, its first two moments are

     $`\mathbb{E}[U_Y]=0`$ and $`\mathbb{E}[U_Y^2]=1.`$

     Therefore,

     $`d(x,l;\theta)=\int_{-\infty}^{\infty}f_Y(u_y,x,l;\theta_Y)\,p_{\mathcal{N}}(u_y;0,1)\,\mathrm{d}u_y`$<br>
     $`=\theta_{\mathrm{quadratic}}\mathbb{E}[U_Y^2]+\theta_{\mathrm{linear}}\mathbb{E}[U_Y]+\theta_xx+\theta_l\sum_i l_i`$<br>
     $`=\theta_{\mathrm{quadratic}}+\theta_xx+\theta_l\sum_i l_i.`$
  3. `test_causal_response_is_jittable_and_differentiable` does as it says. 
  
For `test_causal_response_matches_standard_normal_moments(...)`, I used `n_points=1,000,000`, adjusted `rtol`  because the errors were too large and added batching to overcome memory issues. 
It works relatively okay at 1,000,000 but after further investigation that just seems to be a local minimum. I have included a graph of `n_points` vs largest absolute error below. 

Also, I was getting inf errors until I started using 64 bit jax. 
  
output (1)


src/causalprog/graph/ricardo.py
```diff
@@ -187,3 +187,101 @@ def _r(
...
+    =
+    \int_{-\infty}^{\infty}
+    f_Y(u_y, x, l; \theta_Y)
+    p_N(u_y)
+    \,\mathrm{d}u_y.
+    $$
+
+    The returned callable has signature ``d(xl, model_params)``, where
+    ``xl`` contains the fixed values of ``x`` and ``l``. The latent variable
+    ``u_y`` is supplied internally by the quadrature rule.
```
willGraham01:
```suggestion
    The returned callable has signature `d(xl, model_params)`, where
    `xl` contains the fixed values of `x` and `l`. The latent variable
    `u_y` is supplied internally by the quadrature rule.
```

Minor nitpick but since we're using mkdocs and not sphinx, inline code only needs a single backtick rather than double backticks.

src/causalprog/graph/ricardo.py
```diff
@@ -187,3 +187,101 @@ def _r(
...
+    ) -> float | NDArray:
+        """Evaluate the outcome function at one quadrature point."""
+        return f_y(
+            {
+                "u_y": u_y,
+                "x": xl["x"],
+                "l": xl["l"],
+            },
+            model_params["theta_y"],
+        )
```
willGraham01:
```suggestion
    def _integrand(
        u_y: float,
        xl: dict[str, float | NDArray],
        model_params: dict[str, ModelParam],
    ) -> float | NDArray:
        """Evaluate the outcome function at one quadrature point."""
        return node_y.compute(
            {
                "u_y": u_y,
                "x": xl["x"],
                "l": xl["l"],
            },
            model_params["theta_y"],
        )
```

`f_y` is kind of redundant here (though I imagine it likely helped to have it when first writing the method). We need `_integrand` of course so that we can pass into the quadrature rule, but I think we can save ourselves a few lines otherwise.

src/causalprog/graph/ricardo.py
```diff
@@ -187,3 +187,101 @@ def _r(
         )
 
     return _r
+
+
+def build_causal_response_function(
+    graph: Graph,
+    quadrature: QuadratureMethod,
+) -> Callable:
```
willGraham01:
```suggestion
) -> MLPAlias:
```

`MLPAlias` is kind of misleading and I should probably update it's name, but it will work as the return typehint here.

Now that I mention it, the same change should be made for `build_regression_function`'s typehint too, so feel free to do that while you're at it 😅

tests/test_integration/_helpers.py
```diff
@@ -0,0 +1,30 @@
+from collections.abc import Iterable
```
willGraham01: To save on adding to the `_helpers.py` file (keeping in mind #158), I recommend moving this function into `tests/fixtures/jax.py`. You can then turn it into a fixture via:

```python
@pytest.fixture
def vectorise_over_dict_args() -> Callable[Concatenate[MLPAlias, ...], MLPAlias]:
    """Vectorise a pure function of dictionary arguments across the dictionary keys.

    This is essentially a wrapper around iterative applications of `jax.vmap` with the
    appropriate `in_axes` specified. The net effect is that if the input `f` was
    being called with a dictionary argument, whose keys were scalar-valued, the returned
    function can be called with the same dictionary argument whose keys are
    vector-valued, and returns a vector-valued output.

    Note that all vmap-ing is done along axis 0. If you want to pass in vector-values
    for some of the dictionary key inputs, ensure that they are aligned along the
    correct axis (each _row_ should be one value of the input).
    """
    def _inner(f: MLPAlias, *dict_keys: Iterable[str]) -> MLPAlias:
        vec_f = f
        all_keys = [key for key_list in dict_keys for key in key_list]
        for key in all_keys:
            vec_f = jax.vmap(
                vec_f,
                in_axes=tuple(
                    {k: None if k != key else 0 for k in arg_keys} for arg_keys in dict_keys
                ),
            )
        return vec_f

    return _inner
```

If you want to use it in a test, just add it to the test arguments like you would any other fixture.

(`Concatenate` is importable from `typing`)

tests/test_integration/test_ricardo/test_causal_response_function.py
```diff
@@ -0,0 +1,279 @@
...
+
+    For U_Y ~ N(0, 1),
+
+    E[U_Y] = 0
+    and
+    E[U_Y**2] = 1.
+
+    The Cartesian product of input values is evaluated in batches to limit
+    peak memory usage.
+    """
```
willGraham01, line 131:
```suggestion
    r"""Compare d against known moments of a standard-normal latent variable.

    For $U_Y \sim \mathcal{N}(0, 1)$,
    
    $$
    \mathcal{E}[U_Y] = 0,
    \quad\text{ and }\quad
    \mathcal{E}{U_Y^2} = 1.
    $$
    
    The Cartesian product of input values is evaluated in batches to limit
    peak memory usage.
    """
```

$\LaTeX$-ifying

tests/test_integration/test_ricardo_causal_response_function.py
```diff
@@ -0,0 +1,279 @@
...
+    ) -> jax.Array:
+        return d(xl, {"theta_y": theta_y})
+
+    def d_analytic(
+        xl: dict,
+        theta_y: dict,
+    ) -> jax.Array:
+        return theta_y["x_scale"] * jnp.exp(-(xl["x"] ** 2)) + theta_y[
+            "l_scale"
+        ] * jnp.sum(xl["l"])
```
willGraham01:
```suggestion
    d_analytic = f_y
```

You could even drop defining `d_analytic` everywhere and just use `f_y` in it's place. Even though `f_y` takes a variable called `u_yxl` as it's first argument, Python doesn't actually care if we only pass in `xl` since we're assuming independence from `u_y` in this test.

tests/test_integration/test_ricardo_causal_response_function.py
```diff
@@ -0,0 +1,279 @@
...
+    analytic_grid = vectorise_over_dict_args(
+        d_analytic,
+        xl.keys(),
+        theta_y.keys(),
+    )
+
+    actual = d_grid(xl, theta_y)
+    expected = analytic_grid(xl, theta_y)
+
+    assert jnp.allclose(actual, expected)
```
willGraham01:
```suggestion
    d_from_theta_y = vectorise_over_dict_args(
        d_from_theta_y,
        xl.keys(),
        theta_y.keys(),
    )
    d_analytic = vectorise_over_dict_args(
        d_analytic,
        xl.keys(),
        theta_y.keys(),
    )

    assert jnp.allclose(d_from_theta_y(xl, theta_y), d_analytic(xl, theta_y))
```

Just a style comment to remove a few lines and extra variable names.

tests/test_integration/test_ricardo_regression_function.py
```diff
@@ -67,8 +42,9 @@ def test_fy_independent_of_uy(
     f_m: MLPAlias = lambda czl, theta_r: czl["c"] * theta_r,
     f_ux: MLPAlias = lambda xzl, theta_x: xzl["x"] * theta_x[0],
     f_y: MLPAlias = lambda xu_y, theta_y: theta_y[0] * jnp.exp(-(xu_y["x"] ** 2)),
-    r_analytic: MLPAlias = lambda xzl, theta: theta["theta_y"][0]
-    * jnp.exp(-(xzl["x"] ** 2)),
+    r_analytic: MLPAlias = lambda xzl, theta: (
+        theta["theta_y"][0] * jnp.exp(-(xzl["x"] ** 2))
+    ),
 ) -> None:
```
willGraham01, line 16: Just a note here that if you make `vectorise_over_dict_args` a fixture, you'll need to include it in the argument list here too

tests/test_integration/test_ricardo/test_causal_response_function.py
```diff
@@ -0,0 +1,279 @@
...
+
+def test_causal_response_matches_standard_normal_moments(
+    rng_key: jax.Array,
+    jax_enable_x64,  # noqa: ARG001
+    n_points: int = 1_000_000,
+    n_eval_pts_per_dim: int = 4,
+    eval_range: tuple[float, float] = (-10.0, 10.0),
+    grid_batch_size: int = 64,
+    atol: float = 0.0,
+    rtol: float = 0.01,
```
willGraham01, line 118: The `rtol` seems quite big given the number of sample points `n_points`. 1 million sample points should be more than enough to perform the integral to suitable accuracy. Though from your graph it looks like this is genuinely the performance our MC integrator are giving us... though I wonder if this is a floating-point-arithmetic thing (if you were getting `inf` errors beforehand, it feels like it could be some kind of under/overflow).

We should look into this if we find out things can't be made tighter. Is this just a case that there's some small region of the grid where things are terrible (and in other places a tighter `rtol` would be fine) or a case that it's genuinely bad everywhere?

willGraham01 (changes requested):
Just some tidying for the test suite but otherwise great job (especially on finding two actually meaningful cases to test that also aren't too complex)! There is one test though where the relative errors seem quite loose.

Also, don't forget to change the PR title (it's currently using your branch name).

tests/test_integration/test_ricardo/test_causal_response_function.py
willGraham01: Now referenced in #170

willGraham01 (approved):

Issue #164, opened 2026-07-22
Initial guess provided to `learn_initialiser` ?


willGraham01:
> Just wondering if we can do something about the initial guess? 
> 
> Currently we have `initial_guess` as a required arg to `stochastic_gradient_descent`, but for this function we have have to include it in the optional `optimiser_args` kwarg here. If we leave it empty then things will break. 
> 
> Wondering if the best solution is to have `initial_guess` be optional for `stochastic_gradient_descent` and have `stochastic_gradient_descent` pick a random start point when `initial_guess` is left empty. 
> 
> Anyway, not important functionality-wise so not important to deal with for the time being.  

 _Originally posted by @samjmolyneux in [#160](https://github.com/UCL/causalprog/pull/160/changes#r3623382312)_

We should standardise how we treat the initial guess for our solvers, whilst also ensuring that external solvers (or at least, `optax` solvers that are our direct dependency) can also be used here. Solutions I foresee:

- Have `initial_guess` be a positional (potentially optional) argument to `learn_initialiser`. Have the `optimiser` input argument to `learn_initialiser` specify that the initial guess must be the 2nd positional argument that the `optimiser` takes. Use this then "guaranteed" format when calling the `optimiser` in the method.
- Have the solvers / optimisers that we provide use a suitable initial guess if one is not provided. This is harder, since coming up with a suitable initial guess might be tricky (particularly since we'd still need to know the format of the `PyTree` object that we're optimising over).

PR #165, opened 2026-07-23
`learn_initialiser` needs to return the function `B`

willGraham01:
Relates to #164 |

#160 introduced the `learn_initialiser` method, but that method returned the arg-min of the objective function $B$ rather than the function $B$ itself. However we will require access to the function $B$ when we come to determine the bounds for the causal response function, and need to use $B$ in the constraints of this problem.

As such, `learn_initaliser` has been renamed to `build_loss_function`, and now just returns the loss function $B$ directly, without attempting to determine the arg-min. This effectively bypasses the issue in #164 since we're now delegating the task of finding the arg-min to the user one they build this function.

This change also makes #133 slightly easier to handle, since the form for $B$ can itself now be an input to `build_loss_function` if we wanted to go that route (with the default being our current sum-of-squares-residual).

### DIFF
--- a/src/causalprog/graph/ricardo.py
+++ b/src/causalprog/graph/ricardo.py
@@ -250,10 +250,16 @@ def build_learn_initialiser(
     It is only necessary to specify arrays that are not mapping over their `0`th axes in
     `evaluation_points_axes_mapping`.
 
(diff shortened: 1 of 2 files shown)


willGraham01:
GitHub pull requests playing up - going to close and re-open

PR #166, opened 2026-07-23
`learn_initialiser` needs to return $B$

willGraham01:
Relates to #164 |

#160 introduced the `learn_initialiser` method, but that method returned the arg-min of the objective function $B$ rather than the function $B$ itself. However we will require access to the function $B$ when we come to determine the bounds for the causal response function, and need to use $B$ in the constraints of this problem.

As such, `learn_initaliser` has been renamed to `build_loss_function`, and now just returns the loss function $B$ directly, without attempting to determine the arg-min. This effectively bypasses the issue in #164 since we're now delegating the task of finding the arg-min to the user one they build this function.

This change also makes #133 slightly easier to handle, since the form for $B$ can itself now be an input to `build_loss_function` if we wanted to go that route (with the default being our current sum-of-squares-residual).


src/causalprog/graph/ricardo.py
```diff
@@ -282,11 +274,11 @@ def learn_initialiser(
         None,
     )
     vectorised_r = jax.vmap(r, in_axes=in_axes)
-    n_eval = r_hat_i.shape[0]
+    n_eval = r_hat_i.size
```
samjmolyneux, line 281: With this, I think we are still leaving a risk of broadcasting problems. 

e.g. 
```python
import jax.numpy as jnp

a = jnp.arange(9).reshape(1, 9)
b = jnp.arange(9).reshape(9, 1)
print(a.size)  # 9
print(b.size)  # 9

```
both of which are shapes that could be produced by different models generating `r_hat_i`.  

I liked that ndim=1 check we had, because then the user instantly knows they aren't using it correctly.

samjmolyneux (approved):
I checked all the test cases. Look all good to me. 

I left one comment about the broadcasting thing. I probably still favour just completely removing the risk of it happening by checking that ndim=1 like you did in a previous pr.

src/causalprog/graph/ricardo.py
willGraham01: I've now added this check (`r_hat_i.ndim == 1`) back in at the start of the function. 

Have kept `.size` here since if we know that `r_hat_i.ndim == 1`, then `.size` will give us the number of elements immediately over having to do something like `jnp.maximum(r_hat_i.shape)` or `jnp.prod(r_hat_i.shape)` etc.

PR #169, opened 2026-07-27
Various fixes in `regression_builder`

willGraham01:
PR fixes a few issues with `regression_builder`, spotted after closer inspection of the document we're working from.

- The $v_y$ term should appear as $\sqrt{v_y}$.
  `regression_builder` works by transforming the argument $U_Y \sim \mathcal{N}(m_y, v_y)$ into $s \sim \mathcal{N}(0, 1)$.
  However, there is a typo in the document that says the appropriate transform is $s = (u_y - m_y) / v_y$, whereas it should in fact be $s = (u_y - m_y) / \sqrt{v_y}$.
- The user can now truncate the domain of integration in the `regression_builder` integral should they so wish (if, for example, they have a priori knowledge of where they expect the integrand to be concentrated).
- `jnp.dot` is explicitly called when computing $m_y$, where before it was not.
  This was resulting in element-wise broadcasting rather than the expected dot-product being performed when both $u$ and $m_y$ were vectors of compatible length.


samjmolyneux (commented):
Good spot on the dot product for element-wise multiplication. I completely missed that! 

Also, I'm guessing we'll want to add bounds for `build_causal_response_function` too.

I'm just getting a little confused on the bounds and it might just be because I have never learnt about quadrature or monte-carlo stuff in detail. 

Here is my current understanding, please correct me if I'm wrong:

So for the purposes of this comment, we can think of original expectation as a slightly more complicated version of this

```math
\mathbb{E}[Y]
=
\int_{-\infty}^{\infty}
f_Y(u_y)\,
p_{\mathcal{N}}(u_y;0,1)\,
\mathrm{d}u_y.
```

But if we then restrict the samples to $a < U_Y < b$, then surely the current `UWMCGQuad`
estimates 

```math
\mathbb{E}[Y \mid a < U_Y < b].
```
and not

```math
\int_a^b
f_Y(u_y)\,
p_{\mathcal{N}}(u_y;0,1)\,
\mathrm{d}u_y
```

But if we let

```math
Z_{a,b}
=
\Pr(a < U_Y < b)
```

Then we can get the integral across $[a,b]$ with

```math
\int_a^b
f_Y(u_y)\,
p_{\mathcal{N}}(u_y;0,1)\,
\mathrm{d}u_y
=
Z_{a,b}\,
\mathbb{E}[Y \mid a < U_Y < b].
```

So if we want to approximate the original full expectation by
ignoring the negligible tails the outside $[a,b]$, we would use

```math
\mathbb{E}[Y]
\approx
Z_{a,b}\,
\mathbb{E}[Y \mid a < U_Y < b],
```


So I suppose my questions about this are 

1. Have I understood this correctly?  
2. Do we want the conditional expectation or the integral between [a, b]?
3. If we want the `UWMCGQuad` to return integral, as it currently says it is doing, do we need to multiply the current output by $$Z_{a,b}$$

willGraham01:
> But if we then restrict the samples to a < U Y < b , then surely the current UWMCGQuad
estimates

Providing $a$ and $b$ to `UWMCGQuad` [results in drawing samples from a truncated normal](https://github.com/UCL/causalprog/blob/e40580fb24df4ee228c0072e9c80e150ebb3386d/src/causalprog/quadrature/monte_carlo.py#L66-L71) on $[a, b]$; so what it's doing is computing

```math
\mathbb{E}[f_Y(u, ...) \vert u \sim \mathrm{truncnorm}_{[a,b]}(0, 1)]
```

in all cases, and so we don't need any adjustment pre-factors in front of the integral. 

In general, the user providing $a$ and $b$ to restrict the domain of an integral they're estimating is effectively them saying that

```math
\int_{-\infty}^{\infty} f \approx \int_{a}^{b} f = \int_{-\infty}^{\infty} 1_{[a,b]} f
```

is a good approximation ($1_{[a,b]}$ is the indicator function on $[a,b]$). So in regards to the expectations above, by providing finite $a$ and/or $b$ we're saying that we belive

```math
\mathbb{E}[f] \approx \mathbb{E}[ 1_{[a,b]} f ].
```

Now in the particular case of Monte-Carlo integration (like `UWMCGQuad`), we rely on generating a suitable number of samples _within the domain of the integral_ to then approximate it. If the user requests 1000 samples, we could generate 1000 samples from the standard normal - no problem there. But if the user has _also_ said the above approximation is valid, we might have to throw say $n$ of these samples away since they're not within $[a,b]$. But then we have to re-generate those $n$ samples, of which we might find that $m$ are still outside $[a,b]$, and this cycle continues which doesn't make for nice code :sweat_smile: Conversely, we can't just use the original 1000 - $n$ samples we generated since if $[a,b]$ is _very tight_ we could feasibly generate only 1 or 2 valid sample points! Using `truncnorm` is just a shortcut to ensure we generate 1000 samples that are normally distributed and all lie within $[a,b]$ in one go.

samjmolyneux (commented):
Sorry to be a hassle, but I just still don't get it. 

Which part of my reasoning was wrong?

I agree with

> Providing $a$ and $b$ to `UWMCGQuad` results in drawing samples from a truncated normal on $[a,b]$; so what it's doing is computing
>
> ```math
> \mathbb{E}[f_Y(u,\ldots) \mid u \sim \mathrm{truncnorm}_{[a,b]}(0,1)]
> ```

But then don't currently agree that this follows: 
> so we don't need any adjustment pre-factors in front of the integral. 

Of course I agree with this following bit for normal integrable functions

>In general, the user providing $a$ and $b$ to restrict the domain of an integral they're >estimating is effectively them saying that
>
>```math
>\int_{-\infty}^{\infty} f \approx \int_{a}^{b} f = \int_{-\infty}^{\infty} 1_{[a,b]} f
>```
>
>is a good approximation ($1_{[a,b]}$ is the indicator function on $[a,b]$). So in regards to the expectations above, by providing finite $a$ and/or $b$ we're saying that we belive
>
>```math
>\mathbb{E}[f] \approx \mathbb{E}[ 1_{[a,b]} f ].
>```

But think the subtlety is that drawing from the truncated normal also changes the density.


Let

```math
Z_{a,b} = \mathbb{P}(a < U < b)
```

where $U \sim \mathcal{N}(0,1)$. The density of the corresponding truncated normal is

```math
q(u)
=
\frac{\phi(u)}{Z_{a,b}},
\quad a \lt u \lt b
```

where $\phi$ is the standard-normal density.

Therefore, an unweighted average over samples from the truncated normal estimates

```math
\mathbb{E}_{U \sim \mathrm{truncnorm}_{[a,b]}(0,1)}
[f_Y(U,\ldots)]
=
\int_a^b f_Y(u,\ldots)q(u)\,\mathrm{d}u
=
\frac{1}{Z_{a,b}}
\int_a^b f_Y(u,\ldots)\phi(u)\,\mathrm{d}u.
```

However, the restricted contribution to the original standard-normal expectation is

```math
\int_a^b f_Y(u,\ldots)\phi(u)\,\mathrm{d}u
=
\mathbb{E}
\left[
1_{[a,b]}(U)f_Y(U,\ldots)
\right].
```

Thus,

```math
\int_a^b f_Y(u,\ldots)\phi(u)\,\mathrm{d}u
=
Z_{a,b}
\mathbb{E}_{U \sim \mathrm{truncnorm}_{[a,b]}(0,1)}
[f_Y(U,\ldots)].
```

And there can be cases where

```math
\int_a^b f_Y(u,\ldots)\phi(u)\,\mathrm{d}u
\approx
\int_{-\infty}^{\infty} f_Y(u,\ldots)\phi(u)\,\mathrm{d}u
\qquad
Z_{a,b} \not\approx 1.
```

If I am obviously wrong and you don't want to waste any time writing out a formal explanation you can just save it for the stand up tomorrow.

willGraham01:
@samjmolyneux As we mentioned in the stand-up, you're right and I wasn't factoring in the change in density functional.

Have re-worked the two classes so they should now be consistent (IE, they should actually be doing what they claim to be doing!) but please double-check this now matches your expectations :sweat_smile:

Also added a couple of test cases to see how well the integrals are performing, and one test case that checks the two schemes are related by integrating $f$ in one of them is akin to integrating $f / p_N$ in the other.

src/causalprog/quadrature/monte_carlo.py
```diff
@@ -2,24 +2,34 @@
...
     $$
 
-    where $p_i\in[a,b]$ are $N$ samples drawn from a standard Gaussian.
+    where
+
+    - $T_{[a,b]}$ is the PDF of a truncated normal distribution on $[a,b]$ with mean 0
+        and variance 1,
+    - $x_i\in[a,b]$ are $N$ samples from the truncated normal distribution defined by
+        $T_{[a,b]}$,
+    - $P = \mathbb{P}[a < X < b \vert X \sim \mathcal{N}(0,1)]$.
```
samjmolyneux:
```suggestion
```

src/causalprog/quadrature/monte_carlo.py
```diff
@@ -51,26 +69,38 @@ def points_and_weights(
...
+    - $x_i\in[a,b]$ are $N$ samples from a truncated normal distribution on $[a,b]$,
+    - $P = \mathbb{P}[a < X < b \vert X \sim \mathcal{N}(0,1)]$.
 
-    This is effectively computing $\mathbb{E}[f(X)]$ when $X$ is distributed according
-    to a truncated normal on $[a, b]$ with mean 0 and standard deviation 1. As one
-    would expect, the above rule for integrating $f$ is identical to conducting standard
-    Monte-Carlo integration (with Gaussian importance sampling), but on the integrand
-    $F(x) = f(x)\mathcal{P}(x)$, where $\mathcal{P}$ is the PDF of a
-    (truncated to $[a, b]$) normal distribution.
+    When $a=-\infty$ and $b=\infy$, this effectively computes
```
samjmolyneux:
```suggestion
    When $a=-\infty$ and $b=\infty$, this effectively computes
```

samjmolyneux (approved):
I've checked the code and tests, they all look correct to me. 

Maybe I would be a little tempted to remove `test_uwmcgq_integration_formula`, although it's not harming anyone!

Issue #171, opened 2026-07-28
Documentation Tidy Up before end-of-ARC-involvement


willGraham01:
We should look to tidy up the documentation for the package before ARC's involvement ends. There are a number of things that we should look to do:

- [x] We should add Ricardo's document, outlining the problem that the package now focuses on, to the documentation build.
- [x] Add any examples that we have to the documentation, and possibly the `examples` directory.
- [ ] Add a "state-of-the-package" / next development steps page to record where we got to.
- [x] Add a "how to get started" page that explains the most common use case (setting up Ricardo's problem, explaining what the `compute` etc attributes of nodes are, going through the various builder functions, etc). #187

PR #172, opened 2026-07-28
Add ricardos document to the docs build

willGraham01:
Concerns #171 |

Does what the title says - adds a page in the documentation that conveys the information in Ricardo's document concerning continuous treatment models.

For the diagram, I couldn't see a more elegant and relatively _short_ solution to adding both the `.svg` file and the corresponding `latex` source to the repo. 

Ideally we would hack into the `mkdocs` / GH-pages build to add a setup-step that rendered the latex source into `.svg` and then ran the build to avoid having to commit two files, but I felt like that was a rabbit hole I didn't want to go down here at least. But we can add it to #171 if we want to go this route later.


samjmolyneux (approved):
I think it will be easier for me to make a new branch to add my changes, will approve this pr and then make my changes in a second pr.

PR #174, opened 2026-07-29
Extract reused mlp test functions into fixtures 

### LINKED ISSUE #157: Tidying `mlp` Tests
A couple of purely code-organisational comments that aren't important enough to address in #152, but that we should look to address in the not-too-distant future.

### LINKED ISSUE #158: Moving `_helpers.py` into the `conftest` scope

### LINKED ISSUE #159: Making MLPs Noisy
`test_mlp_forward_pass_calls_layers_in_expected_order` (in `test_mlps.py`) essentially creates a function that tracks when each "part" of each layer in an MLP is called. We're likely going to want to use this in other integration tests, so it would be good to have a fixture that can do this.

To this end, we should add this as a "toggle" option in our `build_mlp` fixture (or specifically, to the `_build_mlp` inner function) to make the returned model "noisy":

```python
def _build_mlp(listener: list | None = None, **overrides):
    # Do what _build_mlp is doing already... but with
    model, state = mlp(**kwargs)

    if listener is not None:
        for block_index, block in enumerate(model.blocks):
            # We can probably do another inner loop here over the attributes,
            # but that's too complex to write off the top of my head in an issue
            block.linear = record_call(
                listener,
                f"block_{block_index}.linear",
                block.linear,
            )
            block.norm = record_call(
                listener,
                f"block_{block_index}.norm",
                block.norm,
            )
            block.activation = record_call(
                listener,
                f"block_{block_index}.activation",
                block.activation,
            )
            block.dropout = record_call(
                listener,
                f"block_{block_index}.dropout",
                block.dropout,
            )
        model.output_layer = record_call(
            listener,
            "output_layer",
            model.output_layer,
        )

    return model, state
```

The intended use case is then

```python
def my_test(build_mlp):
    listener = []

    # If we want a "noisy" MLP whose layer calls are tracked by the `listener` local variable
    model, state = build_mlp(listener, ...)
    # If we don't want anything noisy
    model, state = build_mlp(...)
```

Note that the `record_call` function, currently defined within `test_mlp_forward_pass_calls_layers_in_expected_order`, would also have to be moved out into `fixtures/mlps.py`. But it does not need to be a fixture itself (it can just be a function defined in the file).

samjmolyneux:
Closes #157
Closes #158
Closes #159

- extract `build_mlp(...)` from `tests.test_mlps._helpers.py` and put  it into `tests/fixtures/mlps.py`
- extract the call recording feature of `test_mlp_forward_pass_calls_layers_in_expected_order` into the `build_mlp` fixture. 
- test that the new `build_mlp` fixture produces the same mlp whether it records calls or not.  

I had to add `test_instrumented_mlp_matches_plain_mlp` along with the `_CallRecorder` `nnx.Module` because the original version was affecting the models ability to traverse the graph and thus merge and split correctly because it no longer recognised the orignal wrapped functions as `nnx.Modules` to be optimised.


tests/fixtures/mlps.py
```diff
@@ -0,0 +1,171 @@
...
+) -> Callable[..., Any]:
+    """Wrap a callable while preserving its static/data status."""
+    if isinstance(fn, nnx.Module):
+        return _CallRecorder(listener, name, fn)
+
+    def wrapped(*args: Any, **kwargs: Any) -> Any:
+        listener.append(name)
+        return fn(*args, **kwargs)
+
+    return wrapped
```
willGraham01, line 65: So we need both `_CallRecorder` class and the `wrapped` fall-back because some blocks have multiple callables in them (which we want to monitor), whereas the `output_layer` is itself a `nnx.Module` that is only directly called?

Tad annoying that we have to have two things that look like they do similar things, but I guess it can't be helped if that's the case (unless we want meta-decorators, but the answer to that is almost always no 😅 )

tests/fixtures/mlps.py
```diff
@@ -0,0 +1,171 @@
...
+
+@pytest.fixture
+def build_mlp() -> Callable[..., tuple[FunctionalMLP, nnx.State]]:
+    """Return a builder for an MLP with standard test defaults."""
+    default_kwargs = {
+        "input_dim": 3,
+        "output_dim": 2,
+        "hidden_layers": 3,
+        "hidden_units": 8,
+    }
```
willGraham01, line 76:
```suggestion
@pytest.fixture
def default_mlp_kwargs() -> dict[str, int]:
    """Default kwargs for building MLPs in tests."""
    return {
        "input_dim": 3,
        "output_dim": 2,
        "hidden_layers": 3,
        "hidden_units": 8,
    }

@pytest.fixture
def build_mlp(default_mlp_kwargs: dict[str, int]) -> Callable[..., tuple[FunctionalMLP, nnx.State]]:
    """Return a builder for an MLP with standard test defaults."""
```

This should be equivalent to what you've got, but also expose the `default_mlp_kwargs` as a fixture too that we can fetch / check against in tests (there might be a few cases where we want these values for some reason, so just future-proofing).

tests/test_mlps/test_mlp_fixtures.py
willGraham01, line 1: 👍 some good future-proofing 😁

willGraham01 (approved):
Great stuff.

I don't know what you think / have already considered regarding my `_CallRecorder` and `_make_call_recorder` comment and possible alternatives. But if there isn't anything that would be easy-to-implement that's already jumped to your mind (I can't think of anything particularly simple to get around this off-the-top-of-my-head) feel free to disregard that comment.

tests/fixtures/mlps.py
samjmolyneux: > EDIT: I think some of the explanation below may be wrong. 
>
> EDIT 2: So what I said below actually isn't entirely incorrect. Parameterised layers do need to be wrapped in `nnx.Module` for the reason I have written below. But it's important also that any layer that contains state information is also an `nnx.Module`. So `nnx.Dropout` isn't parameterised, but it does contain state information about the mlp which must be found during graph traversal and so we must wrap it in the `nnx.Module` also. Basically anything that is an `nnx.Module` originally probably has some reason for being one and thus we need to maintain it. 

Apologies, I should have written a more detailed comment.

So in order to be able to split, merge and train properly, the layers that are parameterised need to be `nnx.Module`s. So for `nnx.Linear`, `nnx.Dropout` and some of the norms we need to wrap them in an `nnx.Module` so the traversal can find the parameters. 

But, we don't want to wrap the non-parameterised layers in `nnx.Module` because flax throws an error that there are no parameters. The activation function is not parameterised and the norm sometimes isn't. 

So we check if the original layer is an `nnx.Module` to know if it's parameterised. If it is we wrap our new logging layer version in the `nnx.Module`.

samjmolyneux:
I've added docstrings to `_make_call_recorder` and `_CallRecorder`. 

Will merge if you are happy with them.

willGraham01 (approved):
Yep all looks good - merge when ready 👍

PR #175, opened 2026-07-31
Add penalty and augmented Lagrangian solvers

mscroggs:
- Added `penalty` and `augmented_lagrangian` solvers
- Added test that these two methods can solve $\min(\frac12(x^2+y^2))$ subject to $x\geqslant1+\epsilon$
- Currently the two files for new methods share a lot of common boilerplate code. It would be possible to instead implement them as specialisations of a more general `MinimisationAlgorithm` class, but this is probably ok for now


src/causalprog/solvers/sgd.py
```diff
@@ -19,7 +19,7 @@ def stochastic_gradient_descent(
     obj_fn: Callable[[PyTree], npt.ArrayLike],
     initial_guess: PyTree,
     *,
-    convergence_criteria: Callable[[PyTree, PyTree], npt.ArrayLike] | None = None,
+    convergence_criterion: Callable[[PyTree, PyTree], npt.ArrayLike] | None = None,
```
willGraham01, line 22: (Proper review pending, but just thought I'd flag)

This is causing the `two_normal_example.py` tests to fail because they refer to the old keyword argument by name.

Might be worth a quick `CTRL+F` across the codebase for the old `convergence_criteria` keyword just to make sure everything is caught

src/causalprog/solvers/augmented_lagrangian.py
```diff
@@ -0,0 +1,185 @@
...
+            `IterationResult` as their only argument, and return `None`.
+            These will be called at the end of each iteration of the optimisation
+            procedure.
+
+    Returns:
+        Result of the optimisation procedure.
+
+    """
+    if fn_args is None:
+        fn_args = ()
```
willGraham01: Tuples are immutable so it's OK to provide `()` as a default for `fn_args` if you want to, and not have to check for `None` here.

But either way of doing things is fine. I just prefer 2 less lines of code 😅

src/causalprog/solvers/augmented_lagrangian.py
```diff
@@ -0,0 +1,185 @@
...
+    if bounds_epsilon is None:
+
+        def evaluate_bounds(guess: dict[str, jax.Array]) -> jax.Array:
+            return jnp.maximum(bounds(guess, *fn_args, **fn_kwargs), 0.0)
+    else:
+
+        def evaluate_bounds(guess: dict[str, jax.Array]) -> jax.Array:
+            return jnp.maximum(
+                bounds(guess, *fn_args, **fn_kwargs) - bounds_epsilon, 0.0
+            )
```
willGraham01:
```suggestion
    if bounds_epsilon is None or bounds_epsilon < 0:
        bounds_epsilon = 0.0


    def evaluate_bounds(guess: dict[str, jax.Array]) -> jax.Array:
        return jnp.maximum(
            bounds(guess, *fn_args, **fn_kwargs) - bounds_epsilon, 0.0
        )
```

You could even give `bounds_epsilon` a default value of 0 and remove the `None` option for passing it's value, if you wanted to.

src/causalprog/solvers/augmented_lagrangian.py
```diff
@@ -0,0 +1,185 @@
...
+    obj_fn: Callable[[PyTree], jax.Array],
+    *args,
+    **kwargs,
+) -> SolverResult:
+    """
+    Penalty method maximisation solver.
+
+    Args:
+        obj_fn: Function to minimise
+
```
willGraham01, line 167:
```suggestion
    Penalty method maximisation solver.

    Thin wrapper around `minimise`, that passes the negation of the `obj_fn` to that method.
    See the corresponding function doc-string for argument and keyword argument options.

    Args:
        obj_fn: Function to minimise

```

src/causalprog/solvers/augmented_lagrangian.py
```diff
@@ -0,0 +1,185 @@
...
+    fn_kwargs: dict | None = None,
+    maxiter: int = 10,
+    tolerance: float = 1.0e-8,
+    history_logging_interval: int = -1,
+    callbacks: Callable[[IterationResult], None]
+    | list[Callable[[IterationResult], None]]
+    | None = None,
+) -> SolverResult:
+    """
+    Minimise a function using a penalty method.
```
willGraham01:
```suggestion
    Minimise a function using the Augmented Lagrangian method.
```

Also maybe worth a link to some kind of source material for the AL methods in the docstring body that follows. The [wikipedia article](https://en.wikipedia.org/wiki/Augmented_Lagrangian_method) would probably do.

src/causalprog/solvers/penalty.py
```diff
@@ -0,0 +1,176 @@
+"""Penalty method solvers."""
```
willGraham01, line 1: My comments on the `augmented_lagrangian` file also apply here. Won't repeat myself but any edits you make to one we should probably also adopt in the other.

tests/test_solvers/test_minimisation.py
```diff
@@ -0,0 +1,52 @@
...
+    return jnp.array([1.0 - x["x1"]])
+
+
+@pytest.mark.parametrize("epsilon", [None, 1e-10, 1e-4, 1e-1, 1.0])
+def test_penalty(epsilon):
+    solution = penalty.minimise(
+        f, {"x1": 0.0, "x2": 0.0}, bounds, bounds_epsilon=epsilon
+    ).fn_args
+
+    if epsilon is None:
```
willGraham01: `epsilon = 0` as a default to the method gets around this `if` statement too 😁

tests/test_solvers/test_minimisation.py
```diff
@@ -0,0 +1,52 @@
+import jax
```
willGraham01: Given that `causalprog` has two separate sub-modules for the general `penalty` method and specific `augmented_lagrangian` methods, I'd be inclined to use two files here too just to mirror the package structure.

tests/test_solvers/test_minimisation.py
```diff
@@ -0,0 +1,52 @@
...
+import pytest
+
+from causalprog.solvers import augmented_lagrangian, penalty
+
+
+def f(x: dict[str, jax.Array]) -> jax.Array:
+    return (x["x1"] ** 2 + x["x2"] ** 2) / 2
+
+
+def bounds(x: dict[str, jax.Array]) -> jax.Array:
```
willGraham01: I'd be inclined to have a couple more tests with different constraints, but the same objective function.

- One-sided bound (this one we currently have)
- "Meaningful" constraints on both $x_1$ and $x_2$. Perhaps keep the current $x_1 \geq 1$ but also have $x_2 \geq 2$ say?
- Two-sided constraints on $x_1$ - maybe force $0.5 \leq x_1 \leq 1$ and see if the max/min problem is solved accordingly.
- A "redundant" constraint, like perhaps the current problem but add $x_2 \leq 10$ as a constraint (which shouldn't affect the result).

willGraham01 (changes requested):
Looks like the tests are failing because [of a variable name change](https://github.com/UCL/causalprog/pull/175#discussion_r3689560390) that has upset another test file. But GitHub won't let me comment on files that aren't changed in the PR to make the fix :upside_down_face: 

Once that's fixed though, happy to greenlight. Though I have some suggestions for a few more low-effort test cases we might want to include just to cover bases regarding the passing of `bounds/constraints` to the problem

src/causalprog/solvers/sgd.py
```diff
@@ -118,7 +116,7 @@ def is_converged(x: npt.ArrayLike, dx: npt.ArrayLike) -> bool:
         history_logging_interval=history_logging_interval,
     )
 
-    for current_iter in range(maxiter + 1):
+    for current_iter in range(maxiter):
```
mscroggs, line 119: I removed a +1 here as it looked like sending in (eg) `maxiter=100` before would actrually do 101 iterations. Or is there some reason I'm missing for a +1 here?

tests/test_solvers/test_penalty.py
```diff
@@ -0,0 +1,128 @@
...
+            id="Upper and lower bounds on x1 and x2.",
+        ),
+        pytest.param(
+            lambda x: jnp.array(
+                [2.0 * x["x1"] + x["x2"] - 2.0, -x["x1"], 1.0 - x["x1"] - x["x2"]]
+            ),
+            {"args": {"x1": 0.5, "x2": 0.5}, "value": 0.25},
+            {"args": {"x1": 0.0, "x2": 2.0}, "value": 2.0},
+            id="A triangular fesible region.",
+        ),
```
mscroggs, line 111: These params are duplicated in two files. This isn't great but probably ok to merge and make it better in the future?

willGraham01: I'm going to go with yes. Can employ the "open an issue so we don't forget" failsafe here if we really want to.

willGraham01 (approved):

PR #176, opened 2026-07-31
add Python 3.14 to test workflow

mscroggs:
If we merge this, we should also update the PR merging requirement to require `Tests / tests (3.14)` to pass


tests/test_backend/test_check_variable_length_parameters.py
```diff
@@ -8,44 +8,11 @@
 @pytest.mark.parametrize(
     ("signature", "expected"),
     [
-        pytest.param(
```
willGraham01, line 11: Normally I'm anti-test removal but I think this backend stuff is from way back when we were trying to be general (and TBH, it could just be nuked from the package in clean-up). So not going to complain about this 😁

willGraham01 (approved):

tests/test_backend/test_check_variable_length_parameters.py
mscroggs: Also in 3.14, Signature has been updated to throw exceptions when initialising (ie here) rather than the error we catch later, so this is one of those unusual cases where this test actually needs removing for future Python releases

willGraham01:
@mscroggs going to merge this now, and update the PR/main-running checks accordingly

PR #177, opened 2026-07-31
Allow learning rate to adapt in penalty methods

willGraham01:
Builds on top of #175 by allowing the user to specify how the learning rate of the inner `stochastic_gradient_descent` solve in the penalty and augmented Lagrangian methods evolves with the penalty terms themselves.

Additionally:

- Unified the max/min problems from being two functions into a single function with a keyword-argument to choose which you want to conduct. Default is `min` for both.
- Exposed the solvers at module level.
- Named the solver methods after their actual names (`penalty_method`, `augmented_lagrangian`) rather than just using `minimise` and relying on imports from the specific sub-modules.

TBH We might want to also expose `stochastic_gradient_descent`'s keyword-arguments to both methods too.


mscroggs (approved):

mscroggs:
| TBH We might want to also expose stochastic_gradient_descent's keyword-arguments to both methods too.

I've opened #179 for this. Could leave it as a good first issue for onboarding a future developer

PR #178, opened 2026-07-31
Clean up Ricardo's specification

samjmolyneux:
So my plan wasn't to rewrite the document, I just noticed that there were a few ambiguities and typos that could use clearing up. 

I spent way too much time on this though because I got so confused with the `hidden variables` section and I feel like that might be because there are slightly bigger issues than just typos or ambiguities, so I wasn't really able to clear up the ambiguities there.  

I also haven't filled in the `Parameter and Hyperparameter Summary` section and the `Query bounds on causal response` is still very concise and could use a bit more explanation, but atleast probably fairly obvious to the three of us what was intended. 

RE the `hidden variables` section, an example of one of my confusions is below:   

From 
> Given $(C = c, Z = z, L = l)$, the conditional mean of $(U_X, U_Y)$ is defined to be zero for all $(c, z, l)$. We define the conditional covariance matrix of $U_X$ to be the $n_Z \times n_Z$ identity matrix, and the variance of $U_Y$ to be 1. What is left to be modelled is the conditional cross-covariance between $U_X$ and $U_Y$, 
> 
>$$ \sigma_{czl} \equiv \mathbb E[U_X U_Y \ \vert \ C = c, Z = z, L = l], $$ 

We have
``` math
\left.
\begin{pmatrix}
U_X \\
U_Y
\end{pmatrix}
\,\right|\,
C=c,\ Z=z,\ L=l
\sim
\mathcal{N}\left(
\begin{pmatrix}
\mathbf{0}_{d_X} \\
0
\end{pmatrix},
\begin{pmatrix}
I_{d_X} & \sigma_{czl} \\
\sigma_{czl}^{\top} & 1
\end{pmatrix}.
\right).
```
which implies

```math
C \perp U_X \mid Z,L.
```

Now for the regression equation, originally we had
> ```math
> \begin{align}
> r(x,z,l)
> &:= \mathbb{E}\!\left[
> Y \mid X=x, Z=z, L=l
> \right] \notag \\
> &= \int
> f_Y(u_y,x,l)\,
> p(u_y\mid u,l)\,
> \mathrm{d}u_y \notag \\
> &= \int
> f_Y(u_y,x,l)\,
> \sum_c \pi_{ul}(c)\,
> p_{\mathcal N}(u_y;m_c,v_c)\,
> \mathrm{d}u_y.
> \end{align}
> ```

But the probability on the second line, $p(u_y \vert u, l)$, should be $p(u_y, \vert u, l, z)$. And when we do that surely we get the following:

```math
\begin{align}
r(x,z,l)
&:= \mathbb{E}\!\left[
Y \mid X=x, Z=z, L=l
\right] \notag \\
&= \int
f_Y(u_y,x,l)\,
p_{U_Y\mid U_X,Z,L}(u_y\mid u,z,l)\,
\mathrm{d}u_y \notag \\
&= \int
f_Y(u_y,x,l)\,
\sum_{c=1}^{K}
\mathbb{P}(C=c\mid U_X=u, Z=z,L=l)\,
p_{\mathcal N}(u_y;m_c,v_c)\,
\mathrm{d}u_y, \\
&= \int
f_Y(u_y,x,l)\,
\sum_{c=1}^{K}
\mathbb{P}(C=c\mid Z=z,L=l)\,
p_{\mathcal N}(u_y;m_c,v_c)\,
\mathrm{d}u_y,
\end{align}
```

Unless we are assuming

```math
\mathbb{p}(C=c \mid U_X=u, Z=z, L=l)\\
=
\pi_{ul}(c) = \mathbb{p}(C=c \mid U_X=u, L=l),
```

or equivalently,

```math
C \perp Z \mid U_X,L.
```
Combined with $C \perp U_X \mid Z, L$ that would give us

```math
\mathbb{P}(C=c \mid U_X=u, Z=z, L=l) = \mathbb{P}(C=c \mid L=l)
```

But surely that contradicts 
```math
\pi_{ul} \equiv \mathrm{softmax}(f_\pi(u, l; \theta_\pi))
```


docs/javascripts/mathjax.js
```diff
@@ -0,0 +1,13 @@
+window.MathJax = {
+  tex: {
+    inlineMath: [["\\(", "\\)"]],
+    displayMath: [["\\[", "\\]"]],
+    processEscapes: true,
+    processEnvironments: true,
+    tags: "ams",
+  },
+  options: {
+    ignoreHtmlClass: ".*|",
```
willGraham01, line 10: Just an FYI, if this is so that you can render vertical bars `|` in equation blocks, the macro `\vert` exists for this. `\vert` works with `\left` and `\right` too. Doing so should mean we don't need this file (unless you're also doing something else with this file)

EG 

```latex
\left\| \right\| --> \left\vert \right\vert
```

samjmolyneux: So I ended up looking into this file because the formula tagging wasn't working and apparently I needed to have this file with the `tags: "ams"` bit. 

While doing this I noticed that the `mkdocs.yml` was already trying to reference the file but couldn't find it because we must have forgot to add it from when the project was created. So I went on the mkdocs docs and basically copy and pasted the file they said we were supposed to have added in their tutorial. I didn't look much into what the other the other bits were doing, so I don't mind how they are changed, although I'm noticing now I forgot to add the refresh bit.  https://squidfunk.github.io/mkdocs-material/reference/math

willGraham01: > So I ended up looking into this file because the formula tagging wasn't working and apparently I needed to have this file with the tags: "ams" bit.

Oh in that case, carry on :joy:

docs/theory/continuous-treatments.md
willGraham01:
```suggestion
### Parameter and Hyperparameter Summary

To summarise, this continuous treatment model has the following parameters and hyperparameters.

Hyper-parameters:

- $K$, defining the range of values that the RV $C$ can take.
- Hyperparameters that are used to specify the feed-forward normalising flow $f_X$, relating $X$ to $U_X$.
- $M_{r}$ and $H_{r}$, the hidden layer specifications for the MLP $f_r$.
- $M_{m}$ and $H_{m}$, the hidden layer specifications for the MLP $f_m$.
- $M_{\pi}$ and $H_{\pi}$, the hidden layer specifications for the MLP $f_{\pi}$.
- $M_{Y}$ and $H_{Y}$, the hidden layer specifications for the MLP $f_Y$.
- The quadrature rule that should be applied to evaluate the regression function $r$ and causal response $d$.

Parameters:

- $\theta_{m}$, the weights and biases of the MLP $f_m$.
- $\theta_{r}$, the weights and biases of the MLP $f_r$.
- $\theta_{\pi}$, the weights and biases of the MLP $f_{\pi}$.
- $\theta_{X}$, the parameters specifying the normalising flow $f_X$ and $g = f_X^{-1}$.
- $\theta_{Y}$, the weights and biases of the MLP $f_Y$.

Note that $\theta_X$ will be ["learnt" during the initial learning stage](#learning-and-querying), and held constant thereafter.
So the parameters of the model that will actually be optimised over are

<!-- prettier-ignore -->
\[
\theta := \left(\theta_X, \theta_\pi, \theta_m, \theta_r, \theta_Y\right).
\]
```

docs/theory/continuous-treatments.md
```diff
@@ -137,17 +166,20 @@ FILL ME IN
...
-Let $r_i(\theta)$ be the evaluation of the regression equation $r(x^{(i)}, z^{(i)}, l^{(i)})$ at parameter value $\theta$, as given by \eqref{eq:regression-model}).
-Here we are making explicit that this expression depends on the union of all model parameters
+Let $r_i(\theta)$ be the evaluation of the regression equation $r(x^{(i)}, z^{(i)}, l^{(i)})$ at parameter value $\theta$, as given by $\eqref{eq:regression-model}$.
+Here we are making explicit that this expression depends on all model parameters
 
-$$ \theta := \theta_X \cup \theta_\pi \cup \theta_m \cup \theta_r \cup \theta_Y. $$
+<!-- prettier-ignore -->
+\[
+\theta = \left(\theta_X, \theta_\pi, \theta_m, \theta_r, \theta_Y\right).
+\]
```
willGraham01:
```suggestion
Here we are making explicit that this expression depends on all model parameters $\theta$.
```

If you fill in the section above, $\theta$ is already defined so we can get away with not having the equation here.

willGraham01 (commented):
Thanks for looking this over and catching all the inconsistencies. I can't give you an answer yet regarding the hidden variables section so we'll postpone that decision till after the stand-up.

I've had a go at filling in the hyper-parameters & parameters section, so feel free to check over what I've put in there.

willGraham01 (approved):
👍 Don't forget to copy-paste your working into an issue

PR #181, opened 2026-07-31
Expose inner SGD options in constrained solvers

### LINKED ISSUE #179: Expose `stochastic_gradient_descent`'s keyword-arguments in `penalty_method` and `augmented_lagrangian`
The `penalty_method` and `augmented_lagrangian` minimisation solvers internally use `stochastic_gradient_descent`. It would very likely be useful to users to be able to pass some argument into the outer solve to be passed on to the inner solver.

lntutor:
## Summary

Add an `sgd_kwargs` option to the penalty and augmented Lagrangian solvers so callers can configure the inner stochastic-gradient-descent solve. Internal objective arguments remain protected while options such as learning rate, iteration limit, tolerance, optimiser, callbacks, and logging can be forwarded.

## Tests

```
uv run --extra test pytest tests/test_solvers/test_penalty.py tests/test_solvers/test_augmented_lagrangian.py -q
uv run ruff check src/causalprog/solvers/penalty.py src/causalprog/solvers/augmented_lagrangian.py tests/test_solvers/test_penalty.py tests/test_solvers/test_augmented_lagrangian.py
uv run ruff format --check src/causalprog/solvers/penalty.py src/causalprog/solvers/augmented_lagrangian.py tests/test_solvers/test_penalty.py tests/test_solvers/test_augmented_lagrangian.py
```

Fixes #179

### DIFF
--- a/src/causalprog/solvers/augmented_lagrangian.py
+++ b/src/causalprog/solvers/augmented_lagrangian.py
@@ -25,6 +25,7 @@ def minimise(
     convergence_criterion: Callable[[PyTree, PyTree], jax.Array] | None = None,
     fn_args: tuple = (),
     fn_kwargs: dict | None = None,
+    sgd_kwargs: dict | None = None,
     maxiter: int = 10,
     tolerance: float = 1.0e-8,
     history_logging_interval: int = -1,
@@ -54,6 +55,9 @@ def minimise(
             solutions.
         fn_args: Positional arguments to be passed to `obj_fn`, and held constant.
         fn_kwargs: Keyword arguments to be passed to `obj_fn`, and held constant.
+        sgd_kwargs: Keyword arguments passed to the inner
+            `stochastic_gradient_descent` solve. `fn_args` and `fn_kwargs` are
+            managed internally and cannot be supplied here.
         maxiter: Maximum number of iterations to perform. An error will be reported if
             this number of iterations is exceeded.
         tolerance: `tolerance` used when determining if a minimum has been found.
@@ -71,6 +75,11 @@ def minimise(
     """
     if fn_kwargs is None:
         fn_kwargs = {}
+    if sgd_kwargs is None:
+        sgd_kwargs = {}
+    if {"fn_args", "fn_kwargs"} & sgd_kwargs.keys():
+        msg = "sgd_kwargs cannot contain fn_args or fn_kwargs."
+        raise ValueError(msg)
     if convergence_criterion is None:
         convergence_criterion = lambda a, b: jnp.sqrt(  # noqa: E731
             sum(l2_normsq(b[i] - a[i]) for i in b)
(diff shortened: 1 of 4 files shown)


lntutor:
Current head `ffbd3fb7b970324f3dd0efb901b8859eacf9b058` is mergeable, and there are no unresolved review threads on the current PR state. This looks ready for maintainer review when convenient.

src/causalprog/solvers/augmented_lagrangian.py
```diff
@@ -71,6 +75,11 @@ def minimise(
     """
     if fn_kwargs is None:
         fn_kwargs = {}
+    if sgd_kwargs is None:
+        sgd_kwargs = {}
+    if {"fn_args", "fn_kwargs"} & sgd_kwargs.keys():
```
willGraham01, line 80:
```suggestion
    elif {"fn_args", "fn_kwargs"} & sgd_kwargs.keys():
```

Since we just checked and handled the default case. Additionally, the `learning_rate` argument to `stochastic_gradient_descent` is also reserved by us in this method, so this should be added to this list.

Also whilst the set-comparison followed by implicit boolean cast does work, you could also consider the more verbose

```suggestion
    elif "fn_args" in sgd or "fn_kwargs" in sgd:
```

just for readability.

src/causalprog/solvers/augmented_lagrangian.py
```diff
@@ -112,11 +121,10 @@ def is_converged(x: PyTree, dx: PyTree) -> bool:
...
         previous_solution = current_solution
+        inner_sgd_kwargs = {"learning_rate": 1 / mu, **sgd_kwargs}
+        inner_sgd_kwargs["fn_kwargs"] = {"mu": mu, "lamb": lamb}
         current_solution = stochastic_gradient_descent(
-            objective,
-            current_solution,
-            fn_kwargs={"mu": mu, "lamb": lamb},
-            learning_rate=1 / mu,
+            objective, current_solution, **inner_sgd_kwargs
         ).fn_args
```
willGraham01, line 128:
```suggestion
current_solution = stochastic_gradient_descent(
            objective,
            current_solution,
            fn_kwargs={"mu": mu, "lamb": lamb},
            learning_rate = 1 / mu,
            **inner_sgd_kwargs,
).fn_args
```

I don't think there's a good reason to pack and then immediately unpack a dictionary here. Plus this has the bonus of making it clear what we are forwarding and what we are managing.

Also, the current implementation silently overrides any user-specified "learning_rate" that might be passed in via `inner_sgd_kwargs`. I think this case should be caught explicitly (and have mentioned this above).

src/causalprog/solvers/penalty.py
willGraham01, line 1: My comments on this file are just analogies of those on the `augmented_lagrangian.py` file.

tests/test_solvers/test_augmented_lagrangian.py
```diff
@@ -34,6 +35,38 @@ def test_maximise():
     assert jnp.isclose(max_solution.obj_val, 5.0 - min_solution.obj_val)
 
 
+def test_minimise_forwards_sgd_kwargs(monkeypatch):
```
willGraham01, line 38: We prefer the use [of `pytest-mock`](https://github.com/pytest-dev/pytest-mock/) in our test suite, as opposed to `monkeypatch`.

This also means that you should be able use `mocker.spy` to record the values that `stochastic_gradient_descent` is called with, rather than needing the current implementation of writing a function that dynamically updates a "global" dictionary. Simply running the `augmented_lagrangian` method for one iteration should be enough to check the arguments are forwarded, without needing a "fake sgd" solver.

tests/test_solvers/test_penalty.py
willGraham01, line 1: Analogous comments here as on the `test_augmented_lagrangian.py` file.

willGraham01 (changes requested):
Hi @lntutor, thanks for making your first contribution!

I left a few comments below which I'd like to discuss / address before approving this PR. Please feel free to use the threads below to reply.

mscroggs:
Closing as there are conflicts with current main branch. Please re-open if you are able to resolve these

PR #184, opened 2026-08-03
Remove Matt Graham from project team / author lists

matt-graham:
My name ended up in project team and citation file metadata as I set up the initial repository using the [UCL-ARC/python-tooling](https://github.com/UCL-ARC/python-tooling/) template and it was possible at that point I would have more involvement in project, but I haven't made any further substantive contributions since so I don't think I need to be included as an author.


willGraham01 (approved):
😢

PR #187, opened 2026-08-03
Add notebook illustrating how to work with the Continuous Treatment model we've provided

### LINKED ISSUE #156: Attaching "extra" MLPs
Original comment here: https://github.com/UCL/causalprog/pull/147/changes#r3520651165

Ricardo's particular example contains some MLPs that are involved in joint distributions, and so don't really fit nicely into our graph framework. In particular, the $U_Y$ node in Ricardo's example requires two MLPs $f_m$ and $f_r$ (with associated parameters $\theta_m$ and $\theta_r$ - these can't be attached as the `_compute` attribute since that role belongs to $f_{\pi}$, so we need to attach them in some other way.

For the purposes of the regression function, they're just assumed to be additional attributes on the node. But something more consistent would be to allow nodes to have a dictionary attribute that users can "attach" additional functions to. We'd need to standardise this at node instantiation, and figure out how we want users to interact with it. We'd also need to figure out how to treat the parameters $\theta_m$ and $\theta_r$ that technically won't be part of the graph structure (or chose to bite the bullet and actually make them part of the graph structure...).

willGraham01:
Concerns #171 |

Now that we have #175 merged, we can write an example on how to interact with the continuous treatment model-framework that we've created. Originally this was going to be added as an integration test, but #116 means that it should be detected and run by the test suite just by including it in the `examples` directory.

PR also adds a page to the documentation that goes through the working for the example problem that is being solved. Where appropriate, the workflow discusses the generalisations that the user is expected to make, notably:

- Using `FunctionalMLP`s in place of the deterministic functions.
- Having pre-fitted a model for $f_X$ and thus knowing $g$ and $\theta_X$.
- Having used a suitable method on $\mathcal{D}_{train}$ to obtain the estimate $\hat{r}$.

The example takes about 75 seconds to run from start-to-end on my 4-year-old laptop.

Note that there are 2 placeholder `FIXME`s in the notebook because of #156 still being needed to be resolved. **Edit** with #200 merged into this branch, this PR now resolves #156.


mscroggs (approved):
There's a TODO in the ipynb (Github won't load the ipynb diff, so can't comment on the exact line). I think it's related to #197 so can be fixed later

Image

Otherwise looks good except for the `FIXME` that ruff is angry about.

PR #190, opened 2026-08-04
Documentation for graph classes

mscroggs:
- Added a documentation page aimed at users
- Added a documentation page aimed at developers
- Added some missing documentation of function args (see #183)
- Made the type hints of some functions consistently use `jax.Array` instead of various `np.typing` things (see #155)

~This documentation assumes that #189 will be merged and will need editing if we decide to not merge that.~ #189 was merged

PR #200, opened 2026-08-05
MLP attachments to the treatment model

willGraham01:
Does the bare minimum objective on #156 |

Some housekeeping:

- Move `MLPAlias` and `ModelParams` definitions into the `_types` module.
- Renames the `ricardo` submodule and related test files to `continuous_treatment`.
- Renames the `example_model` function to `continuous_treatment_model`.
- Adjusts the picture of the treatment model to reflect the latest updates, and be consistent with changes in #178. #198 will need to make further adjustments.

Actual changes:

- Continuous treatment model is constructed without `phi_x`. `L` now directly points to `X` and `U_X`, pending #198.
- Tests are updated accordingly with the new format of the graph.
- We were allowing the various compute functions to be supplied as `None` so I made this de-facto OK by giving them default values of `None`.
- `f_r` and `f_m` are now passed to the model builder function directly. They are still just attached to the `U_Y` node as additional attributes, which doesn't fix the underlying problem that we need some kind of joint-distribution-node, but it does hide the fact that we need to do this.
  - Various fixtures and the example on the branch this PR targets now don't have to have the ugly "attachment" step.

---------

Target is #187. Note that this PR fixes the linting issues that are blocking #187, so recommend:

- Reviewing #187 & getting to the point of approval 
- Reviewing this PR
- Merge this PR into #187 
- Merge the resulting #187 into `main`.


examples/continuous_treatments_linear.ipynb
```diff
@@ -34,13 +34,13 @@
     "import jax.numpy as jnp\n",
     "import jax.random as jrn\n",
     "\n",
-    "from causalprog.graph.ricardo import (\n",
+    "from causalprog.graph.continuous_treatment import (\n",
```
mscroggs, line 37: This location will also need updating in the graph docs (#190) once we merge main into this

src/causalprog/graph/continuous_treatment.py
```diff
@@ -1,52 +1,64 @@
...
+    [the documentation](https://github-pages.ucl.ac.uk/causalprog/theory/continuous-treatments/).
 
     Args:
         label: The label of the graph.
-        l_len: The number of entries in the vector data node l.
-        z_len: The number of entries in the vector data node z.
-        k: The maximum value that could be taken by the mixture indicator c.
-        compute_u_x: Compute u_x given the value of c.
-        compute_u_y: Compute u_y given the value of c.
-        compute_phi_x: Compute phi_x given the value of l.
```
mscroggs, line 44: ... and graph docs will also need updating for removal of phi_x

mscroggs (approved):

Issue #201, opened 2026-08-06
Code Triage


willGraham01:
Creating a running issue that is tracking potentially outdated code that we no longer need going forward, and that we should delete from the codebase to keep the package slim.

Entries qualified with a `?` indicate that whilst I don't think we're currently using this code / it is potentially outdated, it might be worth preserving since an updated / similar implementation might be needed in future.

Note that removing some of this code will result in trimming of the test suite too, and the old `two_normal_example` will also likely need to go (along with it's page in the docs, unless we adapt it into an example in the new framework). 

- `_abc`
  - [ ] `backend_agnostic.py`
- `algorithms`
  - [ ] `moments.py` ? -> Might want to re-write this, or write an accompanying alternative, that uses MC integrals
- [ ] `backend`
  - [ ] `_convert_signature.py`
  - [ ] `_typing.py`
- [ ] `causal_problem` ? -> there might be some pieces here that, long-term, we'd want to integrate into some kind of proper "problem framework". But right now, I think we can ditch them. Note that the `two_noormal_example.ipynb` depends on these classes, but that itself is outdated.
  - [ ] `_base_component.py`
  - [ ] `causal_problem.py`
  - [ ] `components.py`
  - [ ] `handlers.py`
- `graph`
  - `node`
    - [ ] `constant.py` ? -> don't know if we're using this anymore. We have `DataNode` that serves the same purpose.
    - [ ] `distribution.py` ? -> don't think we're using this anymore either, given the new setup. Even if we were, I think the new `RandomVariable` nodes could be used for this in a pinch anyway.
  - `graph.py`
    - [ ] The `.model` method on the `Graph` class ? -> was built when we used `DistributionNodes` as our main method of interaction. Is redundant now since this node type is too, and we want to move towards an `algorithm`-focused interface for the graphs in general.
- `utils`
  - [ ] `translator.py`

mscroggs:
> * `constant.py` ? -> don't know if we're using this anymore. We have `DataNode` that serves the same purpose.

Currently, the only difference is constants have a known value when created, but `DataNode` don't. These could definitely be merged into a single class, similar to how RV nodes have their compute function being optional

PR #206, opened 2026-08-06
Merge DataNode and ConstantNode

mscroggs:
These two node types performed very similar functions. This PR merges them into a single node type.

Progress towards #201


docs/users/graph.md
```diff
@@ -79,32 +79,19 @@ final line of this snippet.
...
 
-A `DataNode` is a node that represents a constant value that is not known
-when the node is created. These nodes have one required keyword argument
-that must be passed: a `label` for the node. They can take the `shape` of
-the data that the node represents as an additional keyword argument, with
-the default shape being `()` for a scalar value.
+A `DataNode` is a node that represents a constant value.
+These nodes have one required keyword argument that must be passed: a `label` for the node.
+They can take the `shape` of the data that the node represents as an additional keyword argument, with the default shape being `()` for a scalar value.
+They can also take the `value` of the data as an additional keyword argument if this is known when they are initialised.
```
willGraham01:
```suggestion

`DataNode`s can also take the `value` of the data as an additional keyword argument if this is known when they are initialised, effectively turning the node into a constant value.
Otherwise, their value can be supplied at runtime to algorithms like `evaluate` in the same manner as other nodes, by providing an appropriate key in the dictionary specifying input values.
```

willGraham01 (approved):
Just a suggestion for a couple of lines that indicate what role the `value` attribute has on behaviour

PR #210, opened 2026-08-06
make tests run on all PRs

### LINKED ISSUE #209: CI tests not running
CI is currently not running tests on #202, I think due to the lines:

```yaml
paths-ignore:
    - "**.md"
```

This means the PR can never be merged due to our merge requirements.

Should we remove that requirement from the test config yaml?

mscroggs:
Resolves #209 I hope

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,8 +5,8 @@ on:
     branches:
       - main
   pull_request:
-    paths-ignore:
-      - "**.md"
+    branches:
+      - "**"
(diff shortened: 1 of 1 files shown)


willGraham01 (approved):
For DeepSKA, we came up with another workaround for this which was to have a "checker" job beforehand which would auto-pass if no `.py` files had been edited.

But I don't think digging into that solution is super necessary here (though noting it here for posterity in case we want this in future). Simple fix is best

